PR #14, opened 2022-10-04
Change expected configuration file naming

andersfischernielsen:
This renames the expected repository YAML configuration naming from `<repository-name>.yaml` to `.config/templates.yaml` to make it clearer for teams what the file contents are expected to be. 

This also resolves a duplicate PR creation bug and avoid storing any files to disk when extracting templates.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 .DS_Store
-private-key.pem
 release.zip
 node_modules
 npm-debug.log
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -1,5 +1,4 @@
 .DS_Store
-private-key.pem
```
andersfischernielsen, line 2: This was just redundant since we have a global ignore for any PEM certificates in the repository already 😄

PR #25, opened 2022-10-11
Ensure PR descriptions reflects changes made in PRs

andersfischernielsen:
This fixes a bug where the PR description did not match the changes made in a PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   ],
   "scripts": {
     "build": "tsc",
-    "build:run": "tsc && yarn start",
+    "build:run": "yarn build && yarn start",
     "start": "probot run ./lib/app.js",
     "test": "jest"
   },
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -2,8 +2,8 @@ import { Context, Probot } from 'probot'
 import { parse } from 'yaml'
 import { RepositoryDetails, RepositoryConfiguration } from './types'
 
-export default (app: Probot, context: Context<'push'>) =>
-  async (fileName: string, repository: RepositoryDetails, sha: string) => {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
   ],
   "scripts": {
     "build": "tsc",
-    "build:run": "tsc && yarn start",
+    "build:run": "yarn build && yarn start",
```
andersfischernielsen, line 16: Done to keep everything the same in the run scripts 😄

src/configuration.ts
```diff
@@ -2,8 +2,8 @@ import { Context, Probot } from 'probot'
 import { parse } from 'yaml'
 import { RepositoryDetails, RepositoryConfiguration } from './types'
 
-export default (app: Probot, context: Context<'push'>) =>
-  async (fileName: string, repository: RepositoryDetails, sha: string) => {
+export const determineConfigurationChanges =
+  (app: Probot, context: Context<'push'>) => async (fileName: string, repository: RepositoryDetails, sha: string) => {
```
andersfischernielsen, line 6: We should generally avoid `default` exports, since they're considered bad practice

src/git.ts
```diff
@@ -176,7 +181,8 @@ const generatePullRequestDescription = (version: string, templates: Template[])
   `
 }
 
-export default (app: Probot, context: Context<'push'>) =>
+export const commitFiles =
+  (app: Probot, context: Context<'push'>) =>
```
andersfischernielsen, line 185: We should generally avoid `default` exports, since they're considered bad practice

andersfischernielsen:
Tests are failing generally - we'll fix this in DX-564 😄

PR #30, opened 2022-10-12
Refactor functions to use Octokit directly instead of passing Context parameters

andersfischernielsen:
This will refactor the app logic to take the required `Octokit` type instead of passing the full `Context` around. This refactor also moves the passing of the required `Octokit` type to the end of the function chains to allow preparing work and executing it at a later stage when an instance of Octokit is passed. 

This will make testing easier and make it more clear what the functions actually do based on the parameters they require.


dpotyralski (approved):
LGTM!

PR #31, opened 2022-10-12
Reduce Octokit type across all function parameters

andersfischernielsen:
This uses the [`Pick`](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys) helper type to indicate which parts of Octokit a given function expects to have available for API calls. 

This will:
- make it easier to reason about what a given function does based on the type of the `octokit` parameter
- enable easier mocking of the `octokit` parameter since fewer object properties are present on the type


dpotyralski (approved):
Nice! 🚀

PR #34, opened 2022-10-13
Ensure Jest configuration for testing

andersfischernielsen:
This will:
- Remove invalid test cases
- Add tests for `src/app.ts`


src/templates.ts
```diff
@@ -78,6 +78,13 @@ export const downloadTemplates =
     })) as OctokitResponse<ArrayBuffer>
     console.debug('Fetched release contents.')
 
+    if (typeof contents === 'string') {
+      return {
+        contents: Uint8Array.from(JSON.parse(contents)).buffer,
+        version: release.tag_name,
+      }
+    }
+
```
dpotyralski, line 87: Why did you decide to add this part? Could you help me understand? In what cases we can get String from `octokit.repos.downloadZipballArchive`?

andersfischernielsen: Well we wouldn't - but Nock doesn't allow mocking a HTTP request that returns an ArrayBuffer 🤔 I'll try to have a rethink on how to do that

andersfischernielsen: Fixed!

dpotyralski (approved):
LGTM! 🍭

PR #39, opened 2022-10-17
Prepending headers support

dpotyralski:
This PR:
* Adds support for prepending headers
* Expose pretending header configuration to environment variable


src/templates.ts
```diff
@@ -1,35 +1,58 @@
...
+    //TODO include in the promise all
+    const codeOwners = await extract(loaded, 'CODEOWNERS')(log)
+
     log.debug(`Extracted ${templates.length} ZIP templates.`)
 
-    return templates
+    return {
+      codeOwners: codeOwners,
+      templates: templates,
+    }
```
andersfischernielsen, line 55: When the properties are named the same as the `const`, then you can skip the colon notation 😄 

Like this:
```suggestion
    return {
      codeOwners,
      templates,
    }
```

src/templates.ts
```diff
@@ -87,6 +110,30 @@ export const downloadTemplates =
...
+
+    let prependingHeader: string
+
+    if (typeof process.env.PREPENDING_HEADER_TEMPLATE != 'undefined' && process.env.PREPENDING_HEADER_TEMPLATE) {
+      prependingHeader = `${process.env.PREPENDING_HEADER_TEMPLATE + matchedWithCodeOwner?.owners}\n\n`
+    } else {
+      prependingHeader = `#OWNER: ${matchedWithCodeOwner?.owners}\n\n`
+    }
+
+    return prependingHeader + mustacheRenderedContent
```
andersfischernielsen: I'm a bit allergic to side-effects. I'd prefer if we just return the result right away to not keep track of the `let` assignment above 😄 

```suggestion
    if (process.env.PREPENDING_HEADER_TEMPLATE) {
      return `${process.env.PREPENDING_HEADER_TEMPLATE + matchedWithCodeOwner?.owners}\n\n${mustacheRenderedContent}`
    } else {
      return `#OWNER: ${matchedWithCodeOwner?.owners}\n\n${mustacheRenderedContent}`
    }
```

WDYT?

test/templates.test.ts
```diff
@@ -1,50 +1,75 @@
...
+  values: { [key: string]: string } = { appName: 'expected-app-name' },
+) => ({
+  version: '0.0.3',
+  files: pathConfigurations,
+  values: values,
+})
 
 describe('Template rendering', () => {
-  test('renders basic json and toml template', async () => {
+  test('should renders basic json and toml template', async () => {
```
andersfischernielsen, line 59: Just a minor typo 😄 

```suggestion
  test('should render basic json and toml template', async () => {
```

src/templates.ts
dpotyralski: `if (process.env.PREPENDING_HEADER_TEMPLATE)` 

this is the equivalent of 

`if (typeof process.env.PREPENDING_HEADER_TEMPLATE != 'undefined' && process.env.PREPENDING_HEADER_TEMPLAT`

?

.env.example
```diff
@@ -9,3 +9,9 @@ TEMPLATE_REPOSITORY_NAME=repository
 TEMPLATE_REPOSITORY_OWNER=organization
 
 BRANCHES_TO_PROCESS=master|main
+
+PREPENDING_HEADER_TEMPLATE="# THIS CODE WAS AUTOGENERATED. DO NOT MODIFY THIS FILE DIRECTLY
+                            # THE SOURCE CODE LIVES IN A DIFFERENT REPOSITORY:
+                            #   - xxxxx
+                            # ACCOUNTABILITY:
+                            # STEWARD:
```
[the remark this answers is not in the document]
dpotyralski: That makes sense, good idea! Would you like to see that change in this PR and on the same task, or we could consider a new task in the form of improvement?

src/templates.ts
andersfischernielsen: Also see: https://developer.mozilla.org/en-US/docs/Glossary/Falsy

andersfischernielsen: Not super nice, but it is what it is. You basically want to always use `===` to check the existence of something and `!` to check for something not being present to be somewhat safe

.env.example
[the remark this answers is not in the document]
andersfischernielsen: I'd just do it in this PR 😄

src/templates.ts
```diff
@@ -87,6 +110,27 @@ export const downloadTemplates =
...
+
+    const codeOwnersEntries = parse(codeowners)
+    const matchedWithCodeOwner = matchFile(template.sourcePath, codeOwnersEntries)
+
+    if (!process.env.PREPENDING_HEADER_TEMPLATE) {
+      log.info("Prepending header template not defined, using default.")
+      return `#OWNER: ${matchedWithCodeOwner?.owners}\n\n` + mustacheRenderedContent
+    }
+
+    return `${process.env.PREPENDING_HEADER_TEMPLATE + matchedWithCodeOwner?.owners}\n\n` + mustacheRenderedContent
```
andersfischernielsen, line 131: These can be inlined into one string in a single pass 😄 

```suggestion
    const header = process.env.PREPENDING_HEADER_TEMPLATE
    if (!header) {
      log.info("Prepending header template not defined, using default.")
      return `#OWNER: ${matchedWithCodeOwner?.owners}\n\n${mustacheRenderedContent}`
    }

    return `${header}${matchedWithCodeOwner?.owners}\n\n${mustacheRenderedContent}`
```

andersfischernielsen: This basically says "try to fetch the header, if it's not there, use the default". That way a header is always present and we can just return a single string 👍🏼

.env.example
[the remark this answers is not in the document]
dpotyralski: Cool!

src/templates.ts
dpotyralski: I see your point, however, have some issues with `process.env.PREPENDING_HEADER_TEMPLATE` as it resolved to an empty string which is "true" for both conditions. 🤔 Setting it up to undefined in tests leads to an "undefined" string 😅

dpotyralski: With [commit](https://github.com/pleo-oss/file-distributor/pull/39/commits/a183252431cc0ba5ceb0d1c7231061a1eb191aa5) applied your suggestion but needed to stay with `?` structure instead of `??` to pass tests.

andersfischernielsen: Ahhh, it's because it's the empty string, probably. Makes sense!

.env.example
andersfischernielsen, line 17:
```suggestion
PREPENDING_HEADER_TEMPLATE="# THIS CODE WAS AUTOGENERATED. DO NOT MODIFY THIS FILE DIRECTLY
                            # THE SOURCE CODE LIVES IN A DIFFERENT REPOSITORY:
                            #   - xxxxx
                            # ACCOUNTABILITY:
                            # STEWARD:"
```

Just to make sure other `.env` variables following this aren't seen as part of the string

dpotyralski: Nice catch!

src/templates.ts
```diff
@@ -87,6 +120,31 @@ export const downloadTemplates =
...
+      log.debug(`File extension: ${templateExtension} with not supported comments`)
+      return mustacheRenderedContent
+    }
+
+    const codeOwnersEntries = parse(codeowners)
+    const matchedCodeOwner = matchFile(template.sourcePath, codeOwnersEntries)
+
+    const header = process.env.PREPENDING_HEADER_TEMPLATE
+      ? process.env.PREPENDING_HEADER_TEMPLATE
+      : '#OWNER: {{{stewards}}}'
```
andersfischernielsen, line 137: These are equivalent 😄 

```suggestion
    const header = process.env.PREPENDING_HEADER_TEMPLATE || '#OWNER: {{{stewards}}}'
```

andersfischernielsen (approved):
Nice! 🚀

.env.example
dpotyralski: But! This didn't help. It only works when everything is in one line - playing around with this a bit more.

src/templates.ts
```diff
@@ -87,6 +120,29 @@ export const downloadTemplates =
     }
   }
 
+const enrichWithPrePendingHeader =
+  (mustacheRenderedContent: string, template: Template, codeowners: string) =>
+  (log: Logger): string => {
+    const templateExtension = template.destinationPath.split('.').pop() as string
+    if (!(templateExtension == 'yaml' || templateExtension == 'toml' || templateExtension == 'yml')) {
```
andersfischernielsen, line 127:
```suggestion
    if (!(templateExtension === 'yaml' || templateExtension === 'toml' || templateExtension === 'yml')) {
```

src/templates.ts
```diff
@@ -87,6 +120,29 @@ export const downloadTemplates =
     }
   }
 
+const enrichWithPrePendingHeader =
+  (mustacheRenderedContent: string, template: Template, codeowners: string) =>
+  (log: Logger): string => {
+    const templateExtension = template.destinationPath.split('.').pop() as string
```
andersfischernielsen, line 126: Is this cast necessary? 
```suggestion
    const templateExtension = template.destinationPath.split('.').pop()
```

src/templates.ts
```diff
@@ -1,35 +1,68 @@
...
+        const contents = await extract(loaded, file.source)(log)
+
+        const sourceTemplateExtension = file.source.split('.').pop()
+        const destinationTemplateExtension = file.destination.split('.').pop()
 
-        const text = (await picked?.async('text')) ?? ''
-        const contents = text?.replace(/#{{/gm, '{{')
+        if (sourceTemplateExtension !== destinationTemplateExtension) {
+          log.error('Template configuration seems to be invalid, file extensions mismatch!')
+          throw Error('Template configuration seems to be invalid, file extensions mismatch!')
```
andersfischernielsen, line 44: I don't think we should throw here - execution can continue, but the repository is going to see a weird PR. That shouldn't require the app to exit IMO 😄

andersfischernielsen: Maybe this can be bumped down to a `log.info` too 🤔 Up to you

src/templates.ts
dpotyralski: Don't fully get it, `==` should be enough here, as we want to compare values of those, not necessarily a type, right?

src/templates.ts
dpotyralski: Hmm I could suggest going with warn and skipping the given template, WDYT?

src/templates.ts
andersfischernielsen: `==` is generally very rarely used these days 😄 It’s strict equality, which we’d expect between the strings `”yaml”` and `”yaml”`

src/templates.ts
andersfischernielsen: Niiice 😄

andersfischernielsen (approved):
Nice!

PR #41, opened 2022-10-18
Extract and provide basic repository information for template values

andersfischernielsen:
This will extract basic repository information (the repository name and default branch) for template rendering. 

This means teams do to not have to provide these values and will therefore reduce human error and make repository configuration easier.


dpotyralski:
Let's remove `BRANCHES_TO_PROCESS` from .env.example

andersfischernielsen:
> do we need to keep it

@dpotyralski Currently we only use it for the templates, but I can extend this PR to always just depend on what we get from the GitHub API

 WDYT?

andersfischernielsen:
Converted the PR back to draft since I realised I can simplify the implementation a bit

src/app.ts
```diff
@@ -1,33 +1,58 @@
...
-  return new RegExp(branches)
+  return {
+    owner: login,
+    repo: name,
+  }
 }
 
-const processPushEvent = (branchesToProcess: RegExp) => async (payload: PushEvent, context: Context<'push'>) => {
-  if (!branchesToProcess.test(payload.ref)) return
+const cache = new Map<string, string>()
```
dpotyralski, line 23: Let's have a bit more descriptive name for this `cache`, that would probably make life easier for any future code reader :) 

Maybe `repositoriesDefaultBranchCache` ?

I'm also wondering about form of invalidation of it, as values stored there at some point can be outdated - this is not happing too often when someone changes the default branch, but it's a valid scenario. Is there an event we could try to catch when this happens?

andersfischernielsen: The cache lives in memory only and will be erased on app redeploy - given how rarely the default branch of repositories are changed and the fact that there's no event for updating it, I don't think it's worth building complex cache invalidation just for this edge case. If we hear about invalid default branches due to caching within `pleo-io`, then we should just restart the pod IMO

TL;DR: Let's cross that bridge when we get to it.

I can rename the cache to something slightly more telling if you think it's not visible from the usage what it's caching 👍🏼

dpotyralski: >The cache lives in memory only and will be erased on app redeploy - given how rarely the default branch of repositories are changed and the fact that there's no event for updating it, I don't think it's worth building complex cache invalidation just for this edge case. 

Mhm, get it. I also agree that there is no point in building complex solutions to handle those cases. Was hoping that there is some notification from GH. My point was to expose the potential issue we have here.

> If we hear about invalid default branches due to caching within pleo-io, then we should just restart the pod IMO

However approaching the problem, that we gonna observe the DataDog for those cases and every time there is an error (if any) restarting the pod to refresh the cache is not really a way to go for me. Consider here this additional element of knowledge sharing, that people need to be aware of this step, any new joiner would need to keep in mind that part of the documentation and react properly to that concrete error message - what's worse, from a knowledge sharing perspective, it won't happen often so it would be easy to forget/omit documentation part related to it or pass that information further.

Also, let's think about potential side effects on how the application reacts in that case, probably no files changed from `getCommitFiles` will get back and the app will silently continue, so no error message won't be even present in our logs.

>TL;DR: Let's cross that bridge when we get to it.

From my POV, with this additional cache layer, we somehow already crossed the bridge with solution we are not sure if is needed. Of course, it will limit the API rest calls, but on the other hand updates on `templates.yaml` files inside repositories won't be so often either.

dpotyralski: Would it be okay with you, if we remove this additional cache layer for now?

andersfischernielsen: We’re gonna get rate-limited by the GitHub API if we fetch the default branch from a repository every time we see a `push` event happening on the organisation.

I don’t really see how we can avoid that without a cache 🤔 

Any ideas? 

If not, then we can revert back to the `env` variable for branches to process to avoid the additional GitHub API request

andersfischernielsen: What I’m saying is: it _is_ needed, either with a cache or another static check 😄

dpotyralski (approved):
LGTM!

PR #44, opened 2022-10-21
Fix tests and make probot throw error

marcos-arranz:
Rethrow errors to be captured by Probot, which will log them. 

Change tests that were failing when propagating the errors.


src/app.ts
```diff
@@ -46,8 +46,7 @@ const processPushEvent = async (payload: PushEvent, context: Context<'push'>) =>
       log.info(`See: https://github.com/${repository.owner}/${repository.repo}/pull/${pullRequestNumber}`)
     }
   } catch (e: unknown) {
-    log.error(`Failed to process commit '${payload.after}' with error:`)
-    log.error(e as never)
+    throw new Error(`Failed to process commit '${payload.after}' with error: ${e}`)
```
andersfischernielsen, line 49: This will kill the app afaict 😬 Not something we'd want - it's fine that a repository fails, but we shouldn't prevent the app from running on all other events coming in after a failure



```suggestion
    log.error(`Failed to process commit '${payload.after}' with error:`)
    log.error(e as never)
```

test/app.test.ts
```diff
@@ -211,8 +216,7 @@ describe('Probot Tests', () => {
       },
     }
 
-    await probot.receive(pushEvent as unknown as EmitterWebhookEvent)
-    expect(logOutput.filter(log => log.level === 50)).toHaveLength(2)
+    await expect(probot.receive(pushEvent as unknown as EmitterWebhookEvent)).rejects.toThrow(Error)
```
andersfischernielsen, line 219: I wouldn't expect the entire app to throw for this - it should continue running

src/app.ts
dpotyralski: Hmm, are you sure? I've just tested it locally and it seems to be not the case. Error is thrown and app keeps going and handling other requests.

marcos-arranz: I have tested locally and it keeps logging the error with the event and goes on.

test/app.test.ts
marcos-arranz: in this case it is just the method `receive` the one that rethrows, on the running app the error is eventually captured

dpotyralski (approved):

PR #46, opened 2022-10-21
Change logging to interpolate using pino

marcos-arranz:
Changes interpolation of variables to use [pino one](https://github.com/pinojs/pino/blob/master/docs/api.md#logger), which should be faster as only evaluates if matches the logger level set up. This changes also inject common properties such as repository name, on every log

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -50,14 +50,14 @@ const processPullRequest = async (payload: PullRequestEvent, context: Context<'p
   const { getTemplateDefaultValues } = templates(log, octokit)
 
   const { number, sha, repository } = extractPullRequestInformation(payload)
-  log.info(`Pull request event happened on #${number}`)
+  log.info('Pull request event happened on #%d', number)
 
   const filesChanged = await getFilesChanged(repository, number)
   const configFile = filesChanged.find(filename => filename === configFileName)
 
   if (!configFile) return
 
-  log.debug(`Found repository configuration file: ${configFile}.`)
+  log.debug('Found repository configuration file: %s.', configFile)
 
   const configurationChanges = await determineConfigurationChanges(configFileName, repository, sha)
   const defaultValues = await getTemplateDefaultValues(configurationChanges.version)
@@ -90,14 +90,14 @@ const processPushEvent = async (payload: PushEvent, context: Context<'push'>) =>
   const { combineConfigurations, determineConfigurationChanges } = configuration(log, octokit)
   const { getTemplateDefaultValues, renderTemplates } = templates(log, octokit)
 
-  log.info(`${context.name} event happened on '${payload.ref}'`)
-
   const repository = extractRepositoryInformation(payload)
   const branchRegex = new RegExp(repository.defaultBranch)
 
+  log.info('%s event happened on %s', context.name, payload.ref)
+
   if (!branchRegex.test(payload.ref)) return
 
-  log.info(`Processing changes made to ${repository.owner}/${repository.repo} in ${payload.after}.`)
+  log.info('Processing changes made in commit %s.', payload.after)
 
   const filesChanged = await getCommitFiles(repository, payload.after)
   if (!filesChanged.includes(configFileName)) return
@@ -110,8 +110,8 @@ const processPushEvent = async (payload: PushEvent, context: Context<'push'>) =>
 
   const { version, templates: processed } = await renderTemplates(combined)
   const pullRequestNumber = await commitFiles(repository, version, processed)
-  log.info(`Committed templates to '${repository.owner}/${repository.repo}' in #${pullRequestNumber}`)
-  log.info(`See: https://github.com/${repository.owner}/${repository.repo}/pull/${pullRequestNumber}`)
+  log.info('Committed templates to %s/%s in #%d', repository.owner, repository.repo, pullRequestNumber)
+  log.info('See: https://github.com/%s/%S/pull/%d', repository.owner, repository.repo, pullRequestNumber)
 }
 
 export = async (app: Probot) => {
--- a/src/git.ts
+++ b/src/git.ts
@@ -11,36 +11,35 @@ export const git = (log: Logger, octokit: Pick<OctokitInstance, 'pulls' | 'repos
       ...repository,
       ref: sha,
     })
-    log.debug('Fetched commit:')
-    log.debug(commit)
+    log.debug('Fetched commit: %o', commit)
+
     const {
       data: { files },
     } = commit
 
     const filenames = files?.map(f => f.filename) ?? []
-    log.debug(`Saw changed files in ${sha}:`)
-    log.debug(filenames)
+    log.debug('Saw changed files in commit %s: %s', sha, filenames.join(','))
 
     return filenames
   }
 
   const getFilesChanged = async (repository: RepositoryDetails, pullRequestNumber: number) => {
-    log.debug(`Fetching files changed for PR #${pullRequestNumber}.`)
+    log.debug('Fetching files changed for PR #%d.', pullRequestNumber)
     const { data: filesChanged } = await octokit.pulls.listFiles({
       ...repository,
       pull_number: pullRequestNumber,
     })
     const filenames = filesChanged.map(file => file.filename)
-    log.debug(`Saw files changed in #${pullRequestNumber}:`)
-    log.debug(filenames)
+
+    log.debug('Saw files changed in #%d: %s', pullRequestNumber, filenames.join(','))
     return filenames
   }
 
   const getOrCreateNewBranch = async (repository: RepositoryDetails, baseBranchRef: string) => {
     try {
-      log.debug(`Creating new branch on SHA: '${baseBranchRef}'.`)
+      log.debug('Creating new branch on SHA: %s.', baseBranchRef)
       const newBranch = (await octokit.git.createRef({ ...repository, ref: fullBranchName, sha: baseBranchRef })).data
-      log.debug(`Created new branch with ref: '${newBranch.ref}'.`)
+      log.debug('Created new branch with ref: %s.', newBranch.ref)
 
       const {
         object: { sha },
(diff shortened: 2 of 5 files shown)


src/app.ts
```diff
@@ -32,28 +32,31 @@ const extractRepositoryInformation = (payload: PushEvent) => {
...
+
   const {
     number,
     pull_request: {
       head: { ref, sha },
     },
   } = payload
 
-  log.info(`Pull request event happened on #${number}`)
+  const log = context.log.child({ ...repository, prNumber: number })
```
andersfischernielsen, line 49: What does `log.child` do here? 🤔

src/app.ts
```diff
@@ -85,42 +87,42 @@ const processPullRequest = async (payload: PullRequestEvent, context: Context<'p
...
-      log.debug(`Requested changes for PR #${number} in ${changeRequestId}.`)
+      log.debug('Requested changes for PR #%s in %s.', number, changeRequestId)
     } else {
       const approvedReviewId = await approvePullRequestChanges(repository, number)(log)(octokit)
-      log.debug(`Approved PR #${number} in ${approvedReviewId}.`)
+      log.debug('Approved PR #%d in %d.', number, approvedReviewId)
     }
 
-    log.info(`Validated configuration changes in #${number} with conclusion: ${checkConclusion}.`)
+    log.info('Validated configuration changes with conclusion: %s.', checkConclusion)
```
andersfischernielsen, line 96:
```suggestion
    log.info('Validated configuration changes in #%s with conclusion: %s.', number, checkConclusion)
```

src/app.ts
```diff
@@ -85,42 +87,42 @@ const processPullRequest = async (payload: PullRequestEvent, context: Context<'p
...
   } catch (error) {
-    log.error(`Failed to process PR #${number}' with error:`)
-    log.error(error as never)
-    throw error
+    const typedError = error as Error
+    throw new Error(`Failed to process with error: ${typedError.message}`)
```
andersfischernielsen, line 99: We shouldn't rethrow here as it will take the entire app process down 😬

src/app.ts
```diff
@@ -85,42 +87,42 @@ const processPullRequest = async (payload: PullRequestEvent, context: Context<'p
...
 
 const processPushEvent = async (payload: PushEvent, context: Context<'push'>) => {
   const { octokit } = context
-  const { log } = context
 
-  log.info(`${context.name} event happened on '${payload.ref}'`)
+  const repository = extractRepositoryInformation(payload)
+  const branchRegex = new RegExp(repository.defaultBranch)
 
-  try {
```
andersfischernielsen, line 108: We shouldn't rethrow here as it will take the entire app process down 😬

src/app.ts
```diff
@@ -85,42 +87,42 @@ const processPullRequest = async (payload: PullRequestEvent, context: Context<'p
...
 
     const filesChanged = await getCommitFiles(repository, payload.after)(log)(octokit)
 
     if (filesChanged.includes(configFileName)) {
       const parsed = await determineConfigurationChanges(configFileName, repository, payload.after)(log)(octokit)
       const { version, templates: processed } = await renderTemplates(parsed)(log)(octokit)
       const pullRequestNumber = await commitFiles(repository, version, processed)(log)(octokit)
-      log.info(`Committed templates to '${repository.owner}/${repository.repo}' in #${pullRequestNumber}`)
-      log.info(`See: https://github.com/${repository.owner}/${repository.repo}/pull/${pullRequestNumber}`)
+      log.info('Committed templates to in PR #%s', pullRequestNumber)
```
andersfischernielsen, line 124:
```suggestion
      log.info('Committed templates to '%s/%s' in PR #%s', repository.owner, repository.repo, pullRequestNumber)
```

src/git.ts
```diff
@@ -11,15 +11,14 @@ export const getCommitFiles =
       ...repository,
       ref: sha,
     })
-    log.debug('Fetched commit:')
-    log.debug(commit)
```
andersfischernielsen, line 15: This is an object - how does the Pino logging handle that? I'd rather log the entire object so we can inspect and filter on it in e.g. DataDog

andersfischernielsen: If we wanna do this, then we should create a helper like: 

```typescript
const debugLog = (log: Logger) => (message: string, object: unknown): void => {
  log.debug({msg: message, object})
}
```

and use it: 

```typescript
debugLog(log)('Fetched commit', commit)
```

src/app.ts
marcos-arranz: Since it is logged in every request as an object the repository and the pr and opted for removing here, but let me know if you think it would be more readable adding in the message too.

src/app.ts
marcos-arranz: Same as in the other comment, the repo owner and name is logged every time, but let me know if it would be clearer to add it in the message here too.

src/git.ts
marcos-arranz: Ah, I missed that `commit` is an object, I can change that. Pino supports logging an object directly with `%o` https://github.com/pinojs/pino/blob/master/docs/api.md#logger

src/git.ts
```diff
@@ -11,148 +11,147 @@ export const getCommitFiles =
...
+    (log: Logger) =>
+      async (octokit: Pick<OctokitInstance, 'git'>) => {
+        const templateTree = templates.map(template => ({
+          path: template.destinationPath,
+          mode: '100644',
+          type: 'blob',
+          content: template.contents,
+        }))
+
+        log.debug(`Fetching existing trees from '${treeSha}'.`)
```
dpotyralski, line 81: It seems there is some code not using pino log interpolation.

src/git.ts
```diff
@@ -11,148 +11,147 @@ export const getCommitFiles =
...
+        } = await octokit.git.getTree({ ...repository, tree_sha: treeSha })
+
+        log.debug('Creating git tree with modified templates.')
+        const {
+          data: { sha: createdTreeSha },
+        } = await octokit.git.createTree({
+          ...repository,
+          tree: [...templateTree, ...existingTree] as [],
+        })
+        log.debug(`Created git tree with SHA '${createdTreeSha}'.`)
```
dpotyralski, line 93: One more :)

src/git.ts
```diff
@@ -11,148 +11,147 @@ export const getCommitFiles =
...
+
+        const {
+          data: { sha: newCommitSha },
+        } = await octokit.git.createCommit({
+          ...repository,
+          message: title,
+          tree: createdTreeSha,
+          parents: [currentCommitSha],
+        })
+        log.debug(`Created git commit with SHA '${newCommitSha}'.`)
```
dpotyralski, line 112: +1

src/git.ts
```diff
@@ -11,148 +11,147 @@ export const getCommitFiles =
...
+
+        log.debug('Creating PR.')
+        const created = await octokit.pulls.create({
+          ...repository,
+          title,
+          body: description,
+          head: fullBranchName,
+          base: baseBranch,
+        })
+        log.debug(`Created PR #${created.data.number}.`)
```
dpotyralski, line 131: +1

src/git.ts
```diff
@@ -11,148 +11,147 @@ export const getCommitFiles =
...
-    })
-    log.debug(`Updated PR #${updated.data.number}.`)
-
-    return updated
-  }
+    (log: Logger) =>
+      async (octokit: Pick<OctokitInstance, 'pulls'>) => {
+        const { title, description } = details
+
+        log.debug(`Updating PR #${number}.`)
```
dpotyralski, line 142: +1

src/git.ts
```diff
@@ -11,148 +11,147 @@ export const getCommitFiles =
...
+        const updated = await octokit.pulls.update({
+          ...repository,
+          pull_number: number,
+          head: fullBranchName,
+          body: description,
+          title: title,
+          state: 'open',
+        })
+        log.debug(`Updated PR #${updated.data.number}.`)
```
dpotyralski, line 151: +1

src/git.ts
```diff
@@ -170,53 +169,53 @@ const getExistingPullRequest =
...
-  }
+    (log: Logger) =>
+      async (octokit: Pick<OctokitInstance, 'pulls'>) => {
+        log.debug(`Attempting automerge of PR #${number}.`)
+        const merged = await octokit.pulls.merge({
+          ...repository,
+          pull_number: number,
+          merge_method: 'squash',
+        })
+        log.debug(`Merged PR #${number}.`)
```
dpotyralski, line 180: +1

src/git.ts
```diff
@@ -170,53 +169,53 @@ const getExistingPullRequest =
...
-      pull_number: number,
-      merge_method: 'squash',
-    })
-    log.debug(`Merged PR #${number}.`)
-
-    return merged
-  }
+    (log: Logger) =>
+      async (octokit: Pick<OctokitInstance, 'pulls'>) => {
+        log.debug(`Attempting automerge of PR #${number}.`)
```
dpotyralski, line 174: +1

src/app.ts
```diff
@@ -99,7 +102,7 @@ const processPullRequest = async (payload: PullRequestEvent, context: Context<'p
     )(log)
 
     const configurationConclusion = await updateCheck(checkId, configurationResult, configurationErrors)
-    log.info(`Validated configuration changes in #${number} with conclusion: ${configurationConclusion}.`)
+    log.info('Validated configuration changes in #%d with conclusion: %s.', number, configurationConclusion)
   } catch (error) {
     log.error(`Failed to process PR #${number}' with error:`)
```
dpotyralski, line 107: One more, not being interpolated.

marcos-arranz: yeah, that one I skipped because it's going to be removed by [PR](https://github.com/pleo-oss/file-distributor/pull/53), but thinking it twice maybe is better to change it here and then resolve the conflict.

dpotyralski (approved):

src/app.ts
marcos-arranz: I have removed it since it didn't add much at the moment

src/app.ts
marcos-arranz: Reverted

PR #49, opened 2022-10-21
Improve concurrency

marcos-arranz:
Modify calls to check to return a Promise so they can be combined and called parallel as there are no dependencies between both calls.


dpotyralski (approved):
LGTM!

PR #50, opened 2022-10-24
Set the default template location to '.github/templates.yaml'

andersfischernielsen:
This sets the default template location to `.github/templates.yaml` with a `.env` parameter for setting a custom location after receiving feedback that this would be better for users.


README.md
```diff
@@ -102,6 +102,8 @@ docker run \
   -e WEBHOOK_SECRET=<webhook-secret> \
   -e TEMPLATE_REPOSITORY_NAME=<template-repository-name> \
   -e TEMPLATE_REPOSITORY_OWNER=<template-repository-owner> \
+  -e TEMPLATE_FILE_PATH=.github/templates.yaml \
```
dpotyralski, line 105: 👍

dpotyralski (approved):
LGTM!

PR #52, opened 2022-10-24
Reduce parameters of implementation by exporting built closures

andersfischernielsen:
This greatly reduces the `(log: Logger) => (octokit: ...)` definitions and invocation in the implementation. 

Instead, this PR will refactor all modules to expose a "builder" which - when given the required parameters - returns an object with closures containing the required logging and GitHub API setup. 

This improves readability and simplifies the logic of the main `app` module.


marcos-arranz (approved):
Nice!

src/app.ts
```diff
@@ -32,83 +26,76 @@ const extractRepositoryInformation = (payload: PushEvent) => {
...
-      combined,
-      defaultValueSchema,
-    )(log)
+    const { result, errors } = validateTemplateConfiguration(combined, defaultValueSchema)
 
-    const configurationConclusion = await updateCheck(checkId, configurationResult, configurationErrors)
-    log.info(`Validated configuration changes in #${number} with conclusion: ${configurationConclusion}.`)
+    const conclusion = result ? 'success' : 'failure'
+    const checkInput = { ...repository, sha: sha }
+    const checkId = await createCheckRun(checkInput)
```
marcos-arranz, line 75: There is no need to await as the promise is already resolved inside `createCheckRun`

marcos-arranz: `createCheckRun` should be called at the beginning to give feedback that the check is pending, otherwise we could call `resolveCheckRun` directly

src/app.ts
```diff
@@ -32,83 +26,76 @@ const extractRepositoryInformation = (payload: PushEvent) => {
...
-      sha: sha,
-    }
-    const [checkId, configuration] = await Promise.all([
-      createCheckRun(createCheckInput)(log)(octokit),
-      determineConfigurationChanges(configFileName, repository, sha)(log)(octokit),
-    ])
-
-    const defaultValues = await getTemplateDefaultValues(configuration.version)(log)(octokit)
-    const defaultValueSchema = generateSchema(defaultValues.values)(log)
+    const configuration = await determineConfigurationChanges(configFileName, repository, sha)
```
marcos-arranz, line 64: This one could have been called in parallel to `createCheckRun` to accelerate the execution

src/app.ts
```diff
@@ -32,83 +26,76 @@ const extractRepositoryInformation = (payload: PushEvent) => {
...
+    } else {
+      const approvedReviewId = await approvePullRequestChanges(repository, number)
+      log.debug(`Approved PR #${number} in ${approvedReviewId}.`)
+    }
+
+    log.info(`Validated configuration changes in #${number} with conclusion: ${checkConclusion}.`)
   } catch (error) {
     log.error(`Failed to process PR #${number}' with error:`)
     log.error(error as never)
+    throw error
```
marcos-arranz, line 90: Log and rethrow is not considered a good practice. Probot already los the error rethrown so there is no need to log here

src/checks.ts
```diff
@@ -1,54 +1,54 @@
 import { OctokitInstance, CreateCheckInput, UpdateCheckInput } from './types'
 import { Logger } from 'probot'
 
-export const createCheckRun =
-  (input: CreateCheckInput) => (log: Logger) => async (octokit: Pick<OctokitInstance, 'checks'>) => {
+export const checks = (log: Logger, octokit: Pick<OctokitInstance, 'checks'>) => {
+  const createCheckRun = async (input: CreateCheckInput) => {
```
marcos-arranz, line 5: I believe this `async` is not need since the Promise is already resolved inside the function

PR #54, opened 2022-10-25
Hotfix release fetch and noisy error logging

andersfischernielsen:
This removes template release fetching for `latest`, since it's invalid and allows errors to bubble up for nicer error logging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,13 +20,11 @@
     "format": "prettier --write src/**/*.ts test/**/*.ts && eslint --fix --ext .ts src && eslint --fix --ext .ts test"
   },
   "dependencies": {
-    "@types/node-fetch": "2.6.2",
     "ajv": "8.11.0",
-    "axios": "1.1.2",
     "codeowners-utils": "1.0.2",
+    "dd-trace": "3.6.0",
     "dotenv": "16.0.3",
     "genson-js": "0.0.8",
-    "jsonwebtoken": "8.5.1",
     "jszip": "3.10.1",
     "mustache": "4.2.0",
     "node-stream-zip": "1.15.0",
@@ -38,6 +36,7 @@
     "@octokit/webhooks-types": "6.4.0",
     "@swc/core": "1.3.7",
     "@swc/jest": "0.2.23",
+    "@types/dd-trace": "^0.9.0",
     "@types/jest": "29.1.2",
     "@types/mustache": "4.2.1",
     "@types/node": "18.8.2",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,6 @@
   "dependencies": {
     "ajv": "8.11.0",
     "codeowners-utils": "1.0.2",
-    "dd-trace": "3.6.0",
     "dotenv": "16.0.3",
     "genson-js": "0.0.8",
     "jszip": "3.10.1",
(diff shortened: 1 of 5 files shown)


src/app.ts
```diff
@@ -85,9 +85,7 @@ const processPullRequest = async (payload: PullRequestEvent, context: Context<'p
 
     log.info(`Validated configuration changes in #${number} with conclusion: ${checkConclusion}.`)
   } catch (error) {
-    log.error(`Failed to process PR #${number}' with error:`)
-    log.error(error as never)
-    throw error
+    log.error(error, `Failed to process PR #${number}`)
```
marcos-arranz, line 88: This is going to make tests in `app.test.ts` pass even if they throw an exception

andersfischernielsen: Nope, the test explicitly validates that an error has been logged [here](https://github.com/pleo-oss/file-distributor/pull/54/files#diff-ba6d46dd662f63ed71795ba931c5bfd4260591e76a23126a5956f8e4d0a7d24aR167)

andersfischernielsen: That's why we have the overriden logger, so we can validate the expected error 😄

marcos-arranz: In this test yes, but all tests become more brittle. On PR https://github.com/pleo-oss/file-distributor/pull/44 there were many tests passing in spite of `nok` not matching routes, just because an exception would be thrown and swallowed in the `try` `catch`

andersfischernielsen: Tried to see whether we can spy on the internal `log` of the `Context`, but my `jest` skills aren't good enouhg 🤔 I'll see if I can make the output nice while throwing, then

andersfischernielsen: Done, WDYT?

andersfischernielsen: Given that we already log the PR or commit we're processing, it doesn't really make sense to log and re-throw at all. It is after all a bad design pattern 🥵 

I've remove the global try/catch in the event handlers now - the tests are very strict as a result and we'll still get nice error logs - LMK WYT 🙏🏼

package.json
```diff
@@ -20,13 +20,10 @@
     "format": "prettier --write src/**/*.ts test/**/*.ts && eslint --fix --ext .ts src && eslint --fix --ext .ts test"
   },
   "dependencies": {
-    "@types/node-fetch": "2.6.2",
     "ajv": "8.11.0",
-    "axios": "1.1.2",
     "codeowners-utils": "1.0.2",
     "dotenv": "16.0.3",
     "genson-js": "0.0.8",
-    "jsonwebtoken": "8.5.1",
```
andersfischernielsen, line 29: Cleaned these redundant dependencies up since they are not needed for the app to run and bloat the `node_modules`

src/app.ts
```diff
@@ -53,42 +52,36 @@ const processPullRequest = async (payload: PullRequestEvent, context: Context<'p
   const { number, sha, repository } = extractPullRequestInformation(payload)
   log.info(`Pull request event happened on #${number}`)
 
-  try {
```
andersfischernielsen, line 56: Just removed the try and fixed the indentation - these changes look scarier than they are

src/app.ts
```diff
@@ -99,32 +92,26 @@ const processPushEvent = async (payload: PushEvent, context: Context<'push'>) =>
 
   log.info(`${context.name} event happened on '${payload.ref}'`)
 
-  try {
```
andersfischernielsen, line 102: Just removed the `try` and fixed the indentation - these changes look scarier than they are

src/app.ts
marcos-arranz: there is a little trick in GH to ignore the whitespaces changes in PT. Just adding `?w=1` at the end of the URL ignores whitespaces

marcos-arranz (approved):

PR #55, opened 2022-10-25
Support sending application traces/metrics to DataDog

dpotyralski:
This PR will add support for DataDog's APM feature and instantiate this on startup.


.env.example
```diff
@@ -8,6 +8,9 @@ WEBHOOK_SECRET=secret
 TEMPLATE_REPOSITORY_NAME=repository
 TEMPLATE_REPOSITORY_OWNER=organization
 
+DD_SERVICE=file-distributor
+DD_ENV=staging
```
andersfischernielsen, line 12:
```suggestion
DD_ENV=cicd
```

package.json
```diff
@@ -24,6 +24,7 @@
     "ajv": "8.11.0",
     "axios": "1.1.2",
     "codeowners-utils": "1.0.2",
+    "dd-trace": "^3.6.0",
```
andersfischernielsen, line 27: This dependency isn't pinned - we should pin it 😄 

Also, I'd really like to avoid to add logging-specific packages to the project. We can probably initiate `dd-trace` with `npx` in a `yarn start:tracing` script

src/app.ts
```diff
@@ -1,3 +1,4 @@
+import 'dd-trace/init';
```
andersfischernielsen, line 1: This function is never called, so I don't think this has any effect 🤔 See: https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/nodejs/?tab=containers#adding-the-tracer-in-code

package.json
andersfischernielsen, line 27:
```suggestion
    "dd-trace": "3.6.0",
```

.env.example
```diff
@@ -8,6 +8,9 @@ WEBHOOK_SECRET=secret
 TEMPLATE_REPOSITORY_NAME=repository
 TEMPLATE_REPOSITORY_OWNER=organization
 
+DD_SERVICE=file-distributor
+DD_ENV=cicd
+
```
andersfischernielsen, line 13: This doesn't apply to everyone using the App, so I wouldn't include this here.

```suggestion

```

andersfischernielsen (approved):
LGTM - if we don't include the Pleo-specific `.env` variables in the `.env.example`

erwinw (approved):

PR #60, opened 2022-10-26
Prefer SWC only in tests and CI

andersfischernielsen:
`docker build` doesn't like `swc`, so we only use it for testing and CI to speed these up, but avoid using it for distributed images.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -24,7 +24,7 @@ jobs:
 
       - name: Build
         run: |
-          yarn build
+          yarn build:swc
 
       - name: Lint
         run: |


dpotyralski (approved):

erwinw (approved):
Sadness. Any idea why docker build dislikes swc?

andersfischernielsen:
> Sadness. Any idea why docker build dislikes swc?

@erwinw No, I couldn't figure it out. `swc` builds completely valid output locally, but when building in the Docker image it breaks 🤔 I'm wondering if the Linux distribution of `swc` lacks all the parameters the macOS version does 🤷🏼 It's really strange behaviour

PR #61, opened 2022-10-26
Default configuration moved out from test directory

dpotyralski:


andersfischernielsen (commented):
Unsure what this is for, without any changes it just breaks tests 😄

PR #67, opened 2022-10-27
Create base configuration PRs for repositories where not present

andersfischernielsen:
This will (if enabled) create onboarding PRs in repositories that don't have a template configuration. 

If this PR is closed in a repository, the app will ignore that repository going forward.


src/app.ts
```diff
@@ -130,17 +134,37 @@ const processPushEvent = async (payload: PushEvent, context: Context<'push'>) =>
...
-    await processPushEvent(context.payload as PushEvent, context)
+  app.on('push', (context: Context<'push'>) => {
+    pushFilesToRepository(context.payload as PushEvent, context)
+    addBaseConfiguration(context.payload as PushEvent, context)
   })
 
-  app.on(['pull_request.opened', 'pull_request.synchronize'], async (context: Context<'pull_request'>) => {
-    await processPullRequest(context.payload as PullRequestEvent, context)
+  app.on(['pull_request.opened', 'pull_request.synchronize'], (context: Context<'pull_request'>) => {
+    processPullRequest(context.payload as PullRequestEvent, context)
```
marcos-arranz, line 168: not using `await` wouldn't make return immediately without waiting for the Promise to be resolved?

marcos-arranz:
Correct me if I am wrong, but this would create a PR if the repo hasn't got a template config configured? Do we need for every repo? I am thinking that there are repos, such scripts, that are not interested in having this kind of configuration

src/app.ts
andersfischernielsen: Yeah, given that we're performing more than one action, we don't need to `await`. 
We can just kick off the work for the event and proceed to the next event right away. Eventually the work finishes with a result. 

I realized that since we care about the possible `throw`s within these event handlers, we unfortunately have to `await` the processing of every single event 😞 I've changed the logic accordingly

andersfischernielsen:
> this would create a PR if the repo hasn't got a template config configured

@marcos-arranz Yup, exactly like Renovate. 

The logic does the following: 
- it will create a PR if it sees a `push` to the default branch of a repo that doesn't have the configuration
- the PR states that "if you don't want a config, close this PR)
- if any PR has been created before or any pr created by the app has been closed: do nothing

This mimics Renovate 1:1 when it discovers a new repository 😄

src/app.ts
marcos-arranz: ok, in that case then we can return immediately and capture errors in https://probot.github.io/api/latest/classes/probot.Probot.html#onError. In that way it would process events as they come and process them async. In case of error it would be processed async as well.

marcos-arranz: Anyway we can leave it for another PR, WDYT?

marcos-arranz:
@andersfischernielsen sweet, makes sense

src/app.ts
```diff
@@ -130,17 +134,41 @@ const processPushEvent = async (payload: PushEvent, context: Context<'push'>) =>
...
+    await Promise.all([
+      pushFilesToRepository(context.payload as PushEvent, context),
+      addBaseConfiguration(context.payload as PushEvent, context),
+    ])
   })
 
-  app.on(['pull_request.opened', 'pull_request.synchronize'], async (context: Context<'pull_request'>) => {
-    await processPullRequest(context.payload as PullRequestEvent, context)
+  app.on(['pull_request.opened', 'pull_request.synchronize'], (context: Context<'pull_request'>) => {
+    processPullRequest(context.payload as PullRequestEvent, context)
```
marcos-arranz, line 172: since we are adding `await` on 'push' we should add it here as well

erwinw (approved):
LGTM

README.md
```diff
@@ -147,15 +147,33 @@ docker run \
...
 TEMPLATE_REPOSITORY_OWNER=template-repository-owner
+
+# The header to add to templates in repositories
+PREPENDING_HEADER_TEMPLATE="# THIS CODE WAS AUTOGENERATED."
+
+# The template path prefix to prepend to all string 'files' definitions
+TEMPLATE_PATH_PREFIX=templates/
+
+# Create onboarding PRs in repositories that are not configured
+CREATE_ONBOARDING_PRS=true
```
dpotyralski, line 175: env.example could be enriched as well ;)

src/app.ts
```diff
@@ -130,14 +134,38 @@ const processPushEvent = async (payload: PushEvent, context: Context<'push'>) =>
...
+
+  const { log, octokit } = context
+  const { getDefaultBranchContents, createBaseConfiguration } = git(log, octokit)
+  const { getLatestTemplateVersion } = templates(log, octokit)
+
+  const repository = extractRepositoryInformation(payload)
+  if (!happenedOnDefaultBranch(repository, payload)) return
+
+  const defaultBranchContents = await getDefaultBranchContents(repository, configFileName)
+  if (defaultBranchContents) return
```
dpotyralski, line 148: Let's have here a log debug along with information that config file discovered and skipping base configuration set up.

PR #68, opened 2022-10-27
Use SWC for better CI/test/build performance

andersfischernielsen:
This ensures that the `.swcrc` configuration is present when building container images for distribution. This speeds up builds massively (200-300% depending on CPU). 

Using [SWC](https://swc.rs/) globally for builds also means that we don't need a separate SWC task for CI and testing. These have therefore been removed. 

The resulting container image has been tested locally with expected secrets and runs as expected.


Dockerfile
```diff
@@ -4,6 +4,7 @@ WORKDIR /usr/src/app
 
 COPY yarn.lock package.json tsconfig.json .
 COPY src/ ./src
+COPY .swcrc ./.swcrc
```
erwinw, line 7: Smart.

erwinw (approved):
LGTM

Dockerfile
andersfischernielsen: Haha, I know 😮‍💨

EliasJorgensen (approved):
I
AM
SPEED
🚗

PR #71, opened 2022-11-09
Send comment on PR errors per line

marcos-arranz:
[Linear ticket](https://linear.app/pleo/issue/DX-721/add-errors-on-pr-comments)

This PR adds a comment on each error on the specific line where it failed. In order to do so, a CST ([Concrete Syntax Tree](https://en.wikipedia.org/wiki/Parse_tree)) is used as per https://eemeli.org/yaml/#parser, so all the content is preserved pristine. 

Once the validation is performed a list of errors with an instance path is received. For each error, the path is matched against the tree representation by performing a depth search. When the node is found, it is compared its offset against the line offset to find the particular line.

Example:

Screenshot 2022-11-09 at 11 20 57

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -77,10 +77,10 @@ const processPullRequest = async (payload: PullRequestEvent, context: Context<'p
     log.debug(`Requested changes for PR #${number} in ${changeRequestId}.`)
   }
 
-  if (versionCheckConclusion === 'failure') return
+  if (versionCheckConclusion === 'failure' || !configurationChanges.repositoryConfiguration) return
 
   const { configuration: templateConfiguration, files } = await getTemplateInformation(
-    configurationChanges.repositoryConfiguration.version,
+    configurationChanges.repositoryConfiguration?.version,
   )
   const defaultValueSchema = generateSchema(templateConfiguration.values)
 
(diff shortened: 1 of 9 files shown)


jsfr (approved):
Very nice

src/app.ts
```diff
@@ -73,27 +73,35 @@ const processPullRequest = async (payload: PullRequestEvent, context: Context<'p
   })
 
   if (!versionResult) {
-    const changeRequestId = await requestPullRequestChanges(repository, number, versionErrors)
+    const changeRequestId = await requestPullRequestChanges(repository, number, configFileName, versionErrors)
     log.debug(`Requested changes for PR #${number} in ${changeRequestId}.`)
   }
 
-  if (versionCheckConclusion === 'failure') return
+  if (versionCheckConclusion === 'failure' || !configurationChanges.repositoryConfiguration) return
```
dpotyralski, line 80: Maybe we could put here some log.debug about what is happening?

dpotyralski (approved):
LGTM!

src/app.ts
marcos-arranz: Yeah, makes sense but if you don't mind I will move it to a different PR where we can tackle unexpected scenarios like timeouts, or lack of repository configuration. I had to add it here to avoid complaints of nullable `repositoryConfiguration`

PR #73, opened 2022-11-14
Add line comment on values errors

marcos-arranz:
Adds line comment on values schema and refactors code for pull request event processing.

After the refactoring now it only validates the schema once: 
 - If the version is wrong, that's added to the errors list and validates the current config against the default schema
 - If the version is valid, the default values are fetched from the repo and a complete schema is created using `ajv-merge-patch` extension. The current config is validated against that schema


src/templates.ts
```diff
@@ -85,7 +87,20 @@ export const templates = (log: Logger, octokit: Pick<OctokitInstance, 'repos'>)
     }
 
     log.debug("Fetching templates from '%s/%s'.", templateRepository.owner, templateRepository.repo)
-    const release = await getReleaseFromTag(templateVersion, templateRepository)
+    let release
+    try {
+      release = await getReleaseFromTag(templateVersion, templateRepository)
+    } catch (error) {
+      if (error instanceof RequestError && error.status === 404) {
+        throw new VersionNotFoundError(
```
dpotyralski, line 95: Maybe instead of throwing a new error and later try catching it, we could enrich the method return type to something like:
```
Promise<TemplateInformation | VersionNotFoundError>
```

and handle it later in a way that as it is usually done for Result/Tuple/Either ?

dpotyralski: Now after looking closer to the code, that change could be a bit hard to get. Not sure if we should it now. Anyway, let me know your thoughts around it.

jsfr (approved):
Looks good, but if possible, I'd love to see an example of the code running on a PR to show how it looks like now.

marcos-arranz:
here is some screenshots:
Screenshot 2022-11-16 at 11 51 34
Screenshot 2022-11-16 at 11 51 26
Screenshot 2022-11-16 at 11 50 55

I added you as a collaborator on my private test repo, you can check it [here](https://github.com/marcos-arranz/test-workflow-actions/pull/25)

marcos-arranz:
@jsfr 
here is an example. 
Screenshot 2022-11-16 at 12 16 01



You can check the PR https://github.com/marcos-arranz/test-workflow-actions/pull/25, I've added as a collaborator

src/templates.ts
dpotyralski: >In this case, I've tried to do a demo with Either but there is a lot needed to be changed as Either needs to be bubbled up.

Mhm, all good - I was also expecting that.

dpotyralski (approved):
LGTM!

PR #76, opened 2022-11-16
Simplify code and fix createTreeChanges

marcos-arranz:
This changes the way createTree changes work. A bug has been reported that after changing a variable, the resulting PR didn't contain any change. After debugging the rendering was done fine but the resulting changes didn't contain any file changed. Now it applies the changes to a base commit, and simplifies the code resulting in less calls to GitHub API which were not needed

[Linear ticket](https://linear.app/pleo/issue/DX-726/values-updates-doesnt-trigger-template-generation)

### DIFF
--- a/src/git.ts
+++ b/src/git.ts
@@ -4,6 +4,7 @@ import { OctokitInstance, PRDetails, RepositoryDetails, Template, ValidationErro
 const baseBranchName = 'centralized-templates'
 const reducedBranchName = `heads/${baseBranchName}`
 const fullBranchName = `refs/${reducedBranchName}`
+import { RequestError } from '@octokit/request-error'
 
 export const git = (log: Logger, octokit: Pick<OctokitInstance, 'pulls' | 'repos' | 'git' | 'issues'>) => {
   const getCommitFiles = async (repository: RepositoryDetails, sha: string) => {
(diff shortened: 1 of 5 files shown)


dpotyralski (approved):
LGTM!

src/git.ts
```diff
@@ -35,51 +36,32 @@ export const git = (log: Logger, octokit: Pick<OctokitInstance, 'pulls' | 'repos
...
-
+  const createTreeWithChanges = async (templates: Template[], repository: RepositoryDetails, base_tree: string) => {
     log.debug('Creating git tree with modified templates.')
     const {
       data: { sha: createdTreeSha },
     } = await octokit.git.createTree({
       ...repository,
-      tree: [...templateTree, ...existingTree] as [],
+      base_tree: base_tree,
+      tree: templates.map(template => ({
```
dpotyralski, line 59: Something which at least for me would make it easier to read would be to get back with a variable for templatesTree. But this is not a hard requirement from my end.
```
   const templateTree = templates.map(template => ({
       path: template.destinationPath,
       mode: '100644',
       type: 'blob',
       content: template.contents,
     }))
```

marcos-arranz: Yeah, for me it is easer too, but the problem is that we lose the type information. If it is declared in a different variable TS infers the type an gives, for instance, `mode` the type `string`, meanwhile the real type allowed is an enum:
`mode?: "100644" | "100755" | "040000" | "160000" | "120000";`

so it will complain:
Screenshot 2022-11-17 at 10 01 14

If we erase the type by converting to `[]` then we lose the type warnings:
Screenshot 2022-11-17 at 10 01 28
notice that the mode assigned is not of the ones allowed

If we keep it as it is it does detect the wro
Screenshot 2022-11-17 at 10 01 55
ng `mode`:

dpotyralski: We could go with [this](https://stackoverflow.com/questions/37978528/typescript-type-string-is-not-assignable-to-type).

Screenshot 2022-11-17 at 11 39 10

marcos-arranz: Nice one! Changing it now

src/templates.ts
```diff
@@ -170,24 +170,17 @@ export const templates = (log: Logger, octokit: Pick<OctokitInstance, 'repos'>)
...
 
     const defaults = await extract(loaded, 'defaults.yaml')
     log.debug('Saw default configuration: %o', defaults)
 
-    log.debug('Extracting template files')
     const allFiles = Object.keys(loaded.files)
 
-    log.debug('Parsing default configuration.')
     const parsed = parse(defaults) as RepositoryConfiguration
-    log.debug('Parsed default configuration. %o', parsed)
```
jsfr, line 190: Why did we kill all the debug logs?

jsfr (approved):

src/templates.ts
marcos-arranz: They were redundant, these ones didn't add any information about the internal state:
```
    log.debug('Extracting ZIP contents.')
    log.debug('Extracting default configuration.')
    log.debug('Extracting template files')
    log.debug('Parsing default configuration.')
```
and these ones were already logged by another line:
```
    log.debug("Configuration uses template version '%s'.", version)
``` by `log.debug("Downloading templates with version '%s'.", version)`

and `log.debug('Parsed default configuration. %o', parsed)` by `log.debug('Saw default configuration: %o', defaults)`

It made harder to read the logs

PR #78, opened 2022-11-28
Change start script to inject custom octokit retry options

marcos-arranz:
This PR changes the start script to use a custom `Server`, this is done so custom retry options can be passed to it. The reason to have custom retry options is because it has been observed that some times call to retrieve files from a pull request event received throws a 404 error, due likely an eventual consistency isue. By default octokit-retrier does not retry 404 errors. 

Related: https://github.com/probot/probot/discussions/1768#discussioncomment-4239894

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "scripts": {
     "build": "swc src -d lib -s --copy-files",
     "build:run": "yarn build && yarn start",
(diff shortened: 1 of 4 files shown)


andersfischernielsen (approved):
🔥 A shame we have to duplicate the probot definition, but if that's the only way - LGTM

PR #80, opened 2022-12-05
Capture Unhandled errors and add rerun check listener

marcos-arranz:
This captures unhandled YAML parsing errors. On error it returns an `Either` object, which later on is inspected to populate the validation errors

Testing on private repo:

Screenshot 2022-12-05 at 17 13 17

This PR also allows to re run a failed checked.

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,1 +1,0 @@
-nodeLinker: node-modules
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,0 +1,1 @@
+nodeLinker: node-modules
--- a/package.json
+++ b/package.json
@@ -27,7 +27,6 @@
     "fp-ts": "^2.13.1",
     "genson-js": "0.0.8",
     "jszip": "3.10.1",
-    "monet": "^0.9.3",
     "mustache": "4.2.0",
     "node-stream-zip": "1.15.0",
     "probot": "12.2.8",
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "fp-ts": "^2.13.1",
     "genson-js": "0.0.8",
     "jszip": "3.10.1",
+    "monet": "^0.9.3",
     "mustache": "4.2.0",
     "node-stream-zip": "1.15.0",
     "probot": "12.2.8",
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,15 +1,11 @@
-import { PullRequestEvent, PushEvent } from '@octokit/webhooks-types'
-import { Context, Logger, Probot } from 'probot'
+import { CheckRunRerequestedEvent, PullRequestEvent, PushEvent } from '@octokit/webhooks-types'
+import { Context, Probot } from 'probot'
 import { configuration } from './configuration'
 import { templates } from './templates'
-import { git } from './git'
-import { schemaValidator } from './schema-validator'
-import { checks } from './checks'
 import * as E from 'fp-ts/Either'
-import { pipe } from 'fp-ts/function'
+import { git } from './git'
 import 'dotenv/config'
-import { OctokitInstance, TemplateConfig, ValidationError, VersionNotFoundError } from './types'
-import { YAMLParseError } from 'yaml'
+import { coreValidation } from './core-validation'
 
 const configFileName = process.env['TEMPLATE_FILE_PATH'] ? process.env['TEMPLATE_FILE_PATH'] : '.github/templates.yaml'
 
@@ -53,114 +49,44 @@ const extractPullRequestInformation = (payload: PullRequestEvent) => {
   }
 }
 
-const validateChanges = async (
-  log: Logger,
-  octokit: Pick<OctokitInstance, 'repos'>,
-  configurationChangesOrError: E.Either<YAMLParseError, TemplateConfig>,
-): Promise<ValidationError[]> => {
-  const { getTemplateInformation } = templates(log, octokit)
-  const { validateFiles, validateTemplateConfiguration, generateSchema, mergeSchemaToDefault, getDefaultSchema } =
-    schemaValidator(log)
+const extractCheckRunInformation = (payload: CheckRunRerequestedEvent) => {
+  const {
+    check_run: { pull_requests, id },
+    repository,
+  } = payload
 
-  const { combineConfigurations } = configuration(log, octokit)
-
-  async function validateCorrectYamlChanges(configurationChanges: TemplateConfig) {
-    try {
-      const { configuration: defaultValuesConfiguration, files } = await getTemplateInformation(
-        configurationChanges.repositoryConfiguration.version,
-      )
-
-      const defaultValueSchema = generateSchema(defaultValuesConfiguration.values)
-
-      const combined = combineConfigurations(defaultValuesConfiguration, configurationChanges.repositoryConfiguration)
-
-      const validatedTemplates = validateTemplateConfiguration(
-        {
-          repositoryConfiguration: combined,
-          cstYamlRepresentation: configurationChanges.cstYamlRepresentation,
-        },
-        mergeSchemaToDefault(defaultValueSchema),
-      )
-      const validatedFiles = validateFiles(combined, files)
-
-      return validatedTemplates.errors.concat(validatedFiles.errors)
-    } catch (error) {
-      if (error instanceof VersionNotFoundError) {
-        log.debug('Version for %s/%s:%s has not been found', error.owner, error.name, error.version)
-        const validatedTemplate = validateTemplateConfiguration(configurationChanges, getDefaultSchema())
-        return validatedTemplate.errors.concat({
-          message: `Version could not be found at ${error.owner}/${error.repo}:${error.version}`,
-          line: undefined,
-        })
-      } else {
-        throw error
-      }
-    }
+  if (pull_requests.length != 1) return undefined
+  return {
+    number: pull_requests[0].number,
+    sha: pull_requests[0].head.sha,
+    checkId: id,
+    repository: {
+      owner: repository.owner.login,
+      repo: repository.name,
+      defaultBranch: repository.default_branch,
+    },
   }
+}
 
-  return pipe(
-    configurationChangesOrError,
-    E.match(
-      failure =>
-        Promise.resolve([
-          {
-            message: failure.message,
-            line: failure.linePos?.[0].line,
-          } as ValidationError,
-        ]),
-      success => {
-        return validateCorrectYamlChanges(success)
-      },
-    ),
-  )
+const processCheckRerun = async (payload: CheckRunRerequestedEvent, context: Context<'check_run'>) => {
+  const extracted = extractCheckRunInformation(payload)
+  if (extracted === undefined) return
+  const { number: prNumber, sha, repository, checkId } = extracted
+
+  const { log, octokit } = context
+  const enrichedWithRepoLog = log.child({ owner: repository.owner, repository: repository.repo })
+  const processCheckRun = coreValidation(enrichedWithRepoLog, octokit).processCheckRun
+
+  await processCheckRun({ configFileName, prNumber, repository, sha, previousCheckId: checkId })
 }
 
 const processPullRequest = async (payload: PullRequestEvent, context: Context<'pull_request'>) => {
   const { number: prNumber, sha, repository } = extractPullRequestInformation(payload)
   const { log, octokit } = context
   const enrichedWithRepoLog = log.child({ owner: repository.owner, repository: repository.repo })
+  const processCheckRun = coreValidation(enrichedWithRepoLog, octokit).processCheckRun
 
-  const { getFilesChanged, commentOnPullRequest } = git(enrichedWithRepoLog, octokit)
-
-  const { createCheckRun, resolveCheckRun } = checks(enrichedWithRepoLog, octokit)
-  const { determineConfigurationChanges } = configuration(enrichedWithRepoLog, octokit)
-  const conclusion = (errors: ValidationError[]) => (errors.length > 0 ? 'failure' : 'success')
-
-  log.info('Pull request event happened on #%d', prNumber)
-
-  const filesChanged = await getFilesChanged(repository, prNumber)
-  const configFile = filesChanged.find((filename: string) => filename === configFileName)
-
-  if (!configFile) return
-
-  const checkInput = {
-    ...repository,
-    sha: sha,
-  }
-  const checkId = await createCheckRun({
-    ...repository,
-    sha: sha,
-  })
-
-  log.debug('Found repository configuration file: %s.', configFile)
-
-  const configurationChanges = await determineConfigurationChanges(configFileName, repository, sha)
-
-  const errors = await validateChanges(log, octokit, configurationChanges)
-  const comment = await commentOnPullRequest(repository, prNumber, checkId, conclusion(errors))
-  log.debug(`Submitted comment on PR #%d in %s.`, prNumber, comment)
-
-  const checkConclusion = await resolveCheckRun(
-    {
-      ...checkInput,
-      conclusion: conclusion(errors),
-      checkRunId: checkId,
-      errors,
-    },
-    configFileName,
-  )
-
-  log.info(`Validated configuration changes in #%d with conclusion: %s.`, prNumber, checkConclusion)
+  await processCheckRun({ configFileName, prNumber, repository, sha })
 }
 
 const processPushEvent = async (payload: PushEvent, context: Context<'push'>) => {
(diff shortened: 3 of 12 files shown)


.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
marcos-arranz, line 1: This is added by tool volta which I installed for another project. Do you think it can be added to the project @jsfr @andersfischernielsen ? I have noticed the yarn.lock project was also modified, that's the reason I included it

andersfischernielsen: Are you using Yarn 2? 👀 That'd be the reason this is here afaict 🤔 

But it [looks fine to add](https://yarnpkg.com/configuration/yarnrc#nodeLinker)

dpotyralski (approved):
LGTM!

package.json
```diff
@@ -26,6 +26,7 @@
     "dotenv": "16.0.3",
     "genson-js": "0.0.8",
     "jszip": "3.10.1",
+    "monet": "^0.9.3",
```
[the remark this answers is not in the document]
andersfischernielsen: What's the reason for going with `monet` instead of the battle-tested `fp-ts` here?

src/app.ts
```diff
@@ -53,45 +55,60 @@ const extractPullRequestInformation = (payload: PullRequestEvent) => {
...
+      if (error instanceof VersionNotFoundError) {
+        log.debug('Version for %s/%s:%s has not been found', error.owner, error.name, error.version)
+        const validatedTemplate = validateTemplateConfiguration(configurationChanges, getDefaultSchema())
+        return validatedTemplate.errors.concat({
+          message: `Version could not be found at ${error.owner}/${error.repo}:${error.version}`,
+          line: undefined,
+        })
+      } else {
+        throw error
+      }
```
andersfischernielsen, line 96: Shouldn't we convert this to a function returning an `Either` instead of using both exception handling/passing _and_ the functional `Either` design pattern?

andersfischernielsen: Traditionally it's either the functional programming pattern (using pure functions returning `Either` results) _or_ exceptions. Mixing both of them in one code base seems risky and could be confusing for readers IMO

package.json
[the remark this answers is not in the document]
marcos-arranz: No particular reason as far as I know, @dpotyralski and I talked about using more `Either` than exceptions and this was the library @dpotyralski proposed. Do you recomend move to `fp-ts`?

src/app.ts
dpotyralski: > We had this discussion @dpotyralski and I [here](https://github.com/pleo-oss/file-distributor/pull/73#discussion_r1022929714). In an ideal scenario probably yes, but using `VersionNotFound` with `Either` implies chagning several functions. I am not particularly very annoyed by having both patterns in the same codebase, but let me know your thoughts.

TBH I don't recall proposing a concrete library, it was more of a concept proposal :)

package.json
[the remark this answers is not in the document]
dpotyralski: > No particular reason as far as I know, @dpotyralski and I talked about using more `Either` than exceptions and this was the library @dpotyralski proposed. Do you recomend move to `fp-ts`?

TBH I don't recall proposing a concrete library, it was more of a concept proposal :)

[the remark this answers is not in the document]
marcos-arranz: may be, I don’t remember. In any case, I can move it to `fp-ts` if this is a better option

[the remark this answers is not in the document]
andersfischernielsen: `fp-ts` provides a ton of nice helpers for functional programming, so it'd be preferable to use. It's the most complete FP library out there 😄

andersfischernielsen (approved):

package.json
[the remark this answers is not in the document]
andersfischernielsen: `chain` and `pipe` especially for handling the internal `Either` results

test/configuration.test.ts
```diff
@@ -35,9 +37,41 @@ values:
...
+
+    const { determineConfigurationChanges } = configurationSetup(log, mockedOctokit)
+
+    const repositoryDetails = { repo: 'repository', owner: 'pleo-oss', defaultBranch: 'main' }
+
     const result = await determineConfigurationChanges('', repositoryDetails, '')
 
-    expect(result.repositoryConfiguration).toEqual(expected)
+    expect(E.isLeft(result)).toBeTruthy()
+    if (E.isLeft(result)) {
```
marcos-arranz, line 72: Needed as after this check is possible to use `.left`. Otherwise, TS complains about it.

src/app.ts
```diff
@@ -50,104 +49,42 @@ const extractPullRequestInformation = (payload: PullRequestEvent) => {
...
-  const onlyChangesConfiguration = filesChanged.length === 1 && filesChanged[0] === configFileName
-
-  if (errors.length > 0) {
-    const changeRequestId = await requestPullRequestChanges(repository, prNumber, checkId)
-    log.debug(`Requested changes for PR #%d in %s.`, prNumber, changeRequestId)
-  } else if (onlyChangesConfiguration) {
-    const approvedReviewId = await approvePullRequestChanges(repository, prNumber)
-    log.debug(`Approved PR #%d in %s.`, prNumber, approvedReviewId)
-  }
+  await processCheckRun(enrichedWithRepoLog, octokit, configFileName, prNumber, repository, sha, checkId)
```
andersfischernielsen, line 79: I think we should use the same pattern here for the functions that have been extracted as in other modules. That way we don't pass 7 parameters in a single function and can instead encapsulate the logger and Octokit like so:

```suggestion
//Outside of this function
import { validation } from './core-validation'
  //...
  const { processCheckRun } = validation(enrichedLogger, octokit)
  await processCheckRun(configFileName, prNumber, repository, sha, checkId)
```

I still think 5 parameters is too much (does the function do more than one thing or can we combine the data?), but that can be fixed later

src/app.ts
```diff
@@ -50,104 +49,42 @@ const extractPullRequestInformation = (payload: PullRequestEvent) => {
...
-    },
-    configFileName,
-  )
+const processPullRequest = async (payload: PullRequestEvent, context: Context<'pull_request'>) => {
+  const { number: prNumber, sha, repository } = extractPullRequestInformation(payload)
+  const { log, octokit } = context
+  const enrichedWithRepoLog = log.child({ owner: repository.owner, repository: repository.repo })
 
-  log.info(`Validated configuration changes in #%d with conclusion: %s.`, prNumber, checkConclusion)
+  await processCheckRun(enrichedWithRepoLog, octokit, configFileName, prNumber, repository, sha)
```
andersfischernielsen, line 87: Same feedback as https://github.com/pleo-oss/file-distributor/pull/80/files#r1043298474 for this use of the new function.

src/app.ts
```diff
@@ -205,6 +143,10 @@ export = async (app: Probot) => {
     await processPushEvent(context.payload as PushEvent, context)
   })
 
+  app.on('check_run.rerequested', async (context: Context<'check_run'>) => {
+    await processCheckRerun(context.payload as CheckRunRerequestedEvent, context)
+  })
```
andersfischernielsen, line 148: Is this part of the bugfix or new functionality? I'm slightly confused why this is necessary

src/checks.ts
```diff
@@ -27,9 +27,35 @@ export const checks = (log: Logger, octokit: Pick<OctokitInstance, 'checks'>) =>
   const resolveCheckRun = async (input: UpdateCheckInput, configFilePath?: string) => {
     const { repo, owner, checkRunId, sha, conclusion } = input
 
-    if (input.errors.length > 0) {
+    if (
+      input.checkResolution === CheckResultion.ERROR ||
+      (input.checkResolution === CheckResultion.FAILURE && input.errors.length === 0)
+    ) {
```
andersfischernielsen, line 33: Couldn't we use the existing type that GitHub provides here? We can always make is easier to read and pass around by defining a 

```typescript
export type CheckConclusion =
  | 'action_required'
  | 'cancelled'
  | 'failure'
  | 'neutral'
  | 'success'
  | 'skipped'
  | 'timed_out'
  | null
  ```
  type 😄

File: src/checks.ts [hunk 9]
```diff
@@ -27,9 +27,35 @@ export const checks = (log: Logger, octokit: Pick<OctokitInstance, 'checks'>) =>
...
+        name: 'Configuration validation',
+        check_run_id: checkRunId,
+        status: 'completed',
+        head_sha: sha,
+        conclusion,
+        output: {
+          title: 'Schema validation',
+          summary:
+            'There was an unexpected error running the check. Please try again and if the error persists contact the stewards.',
+        },
```
andersfischernielsen, line 48: I'd extract the `if` and `else` out here to separate contained functions. It's really hard for me to understand everything that's happening in these conditionals

src/configuration.ts
```diff
@@ -45,35 +50,42 @@ export const configuration = (log: Logger, octokit: Pick<OctokitInstance, 'repos
...
+      return E.right({
+        repositoryConfiguration: combinedConfiguration,
+        cstYamlRepresentation: rep,
+      } as TemplateConfig)
+    } catch (error) {
+      if (error instanceof YAMLParseError) {
+        return E.left(error)
+      }
+      throw error
+    }
```
andersfischernielsen, line 88: Why are we still catching and rethrowing exceptions? The idea of `Either` is to map every result or failure to a consistent type-safe structure so we avoid having any `try/catch` across the code 🤔

andersfischernielsen: Instead of `parse`, we can use `parseDocument` which doesn't throw on any YAML errors. Then we can just return `document.errors` if there are any `YAMLError`s 😄

src/core-validation.ts
```diff
@@ -0,0 +1,159 @@
...
+} from './types'
+import { checks } from './checks'
+import { ProbotOctokit } from 'probot'
+import { git } from './git'
+
+const validateChanges = async (
+  log: Logger,
+  octokit: Pick<OctokitInstance, 'repos'>,
+  configurationChangesOrError: E.Either<YAMLParseError, TemplateConfig>,
+): Promise<ValidationError[]> => {
```
andersfischernielsen, line 25: I'd prefer (see https://github.com/pleo-oss/file-distributor/pull/80/files#r1043298474) if we could keep the modules in the app consistent 😄 That way it's easier to understand every module when the structure is the same. 

Functions that take either a logger or an Octokit can be part of a function that takes the logger and Octokit and injects them into the functions that use them. These functions are then returned as a function providing:

```typescript
return {
  validateChanges,
  processCheckRun
}
```

That way we limit the parameter passing and a ton of parameters for every function 😄

src/core-validation.ts
```diff
@@ -0,0 +1,159 @@
...
+      {
+        ...checkInput,
+        conclusion: 'failure',
+        checkRunId: checkId,
+        errors: [],
+        checkResolution: CheckResultion.ERROR,
+      },
+      configFileName,
+    )
+    throw error
```
andersfischernielsen, line 157: Why are we throwing if we're returning an `Either`? 🤔

src/types.ts
```diff
@@ -63,10 +63,17 @@ export interface CreateCheckInput {
   sha: string
 }
 
+export enum CheckResultion {
+  ERROR,
+  FAILURE,
+  SUCCESS,
+}
```
andersfischernielsen, line 70: These could be bound to the `'action_required' | 'cancelled' | 'failure' | 'neutral' | 'success' | 'skipped' | 'timed_out' | null` type provided by GitHub (if we _really_ want an enum here, I doubt it) or be converted to a type exactly like GitHub's to avoid having to map from their type to ours 😄

src/types.ts
```diff
@@ -63,10 +63,17 @@ export interface CreateCheckInput {
...
+
 export type UpdateCheckInput = CreateCheckInput & {
   conclusion: string
   checkRunId: number
   errors: ValidationError[]
+  checkResolution: CheckResultion
```
andersfischernielsen, line 76: Isn't this the same as `conclusion`? 🤔

src/types.ts
```diff
@@ -63,10 +63,17 @@ export interface CreateCheckInput {
   sha: string
 }
 
+export enum CheckResultion {
```
andersfischernielsen, line 66:
```suggestion
export enum CheckResolution {
```

test/app.test.ts
```diff
@@ -106,11 +104,11 @@ describe('Probot Tests', () => {
     }
 
     await probot.receive(pushEvent as unknown as EmitterWebhookEvent)
+    expect(baseNock.isDone()).toBeTruthy()
```
andersfischernielsen, line 107: What does this do? 🤔  Shouldn't Nock already explode by itself for any unresolved mocks?

package.json
```diff
@@ -55,5 +56,9 @@
     "smee-client": "1.2.3",
     "ts-jest": "29.0.3",
     "typescript": "4.8.4"
+  },
+  "volta": {
+    "node": "18.12.1",
+    "yarn": "1.22.19"
```
andersfischernielsen, line 62: Why do we need the Volta configuration again? None of the project relies on Volta for building and distribution 🤔

andersfischernielsen:
Maybe we should split the bugfix into a `patch` PR and add the additional functionality in a `minor` PR. That'd also be making it easier to handle the module, parameter and structure comments, if there are any 😄

marcos-arranz:
I think it makes sense to split it @andersfischernielsen , this PR is quite big and has two different purposes. I am going to address your comments here and send a different PR for the YAML error, so comments are preserved

package.json
marcos-arranz: I installed as per instructions in Grim Reaper https://github.com/pleo-io/grim-reaper/blob/main/package.json#L65. I haven't found any way to tell volta to pin yarn without modifying `package.json`

src/configuration.ts
marcos-arranz: > Instead of parse, we can use parseDocument which doesn't throw on any YAML errors. Then we can just return document.errors if there are any YAMLErrors 😄

`parseDocument` seems to return a different type that cannot be safely casted to `RepositoryConfiguration`

src/app.ts
marcos-arranz: Part of the new functionality. It is necessary because if the check fails because a transient error, like a network error, the user probably wants to re run the check. If we don't implement this then the button on the checks won't do anything.

src/app.ts
```diff
@@ -170,9 +107,10 @@ const processPushEvent = async (payload: PushEvent, context: Context<'push'>) =>
   const filesChanged = await getCommitFiles(repository, payload.after)
   if (!filesChanged.includes(configFileName)) return
 
-  const parsed = await determineConfigurationChanges(configFileName, repository, payload.after)
+  const either = await determineConfigurationChanges(configFileName, repository, payload.after)
```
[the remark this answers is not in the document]
marcos-arranz: `parsed` is used below alreay. I can change `either` to `errorOrParsed`. WDYT?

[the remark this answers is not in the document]
marcos-arranz: Or use pipe as in 

```
return pipe(
    either,
    E.match(
      () => undefined,
      async (parsed: TemplateConfig) => {
        const { configuration: defaultValues } = await getTemplateInformation(parsed.repositoryConfiguration.version)

        const combined = combineConfigurations(defaultValues, parsed.repositoryConfiguration)
        if (!combined) return

        const { version, templates: processed } = await renderTemplates(combined)
        const pullRequestNumber = await commitFilesToPR(repository, version, processed)

        if (!pullRequestNumber) {
          log.info('Commit leads to no changes.')
          log.info('Skipped PR creation.')
        }

        log.info('Committed templates to %s/%s in #%d', repository.owner, repository.repo, pullRequestNumber)
        log.info('See: https://github.com/%s/%s/pull/%d', repository.owner, repository.repo, pullRequestNumber)
      },
    ),
  )
```

but I feel it's a bit overkill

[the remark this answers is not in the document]
marcos-arranz: Also, I am not sure how it works with the `async` modifier

src/checks.ts
marcos-arranz: Unfortunately Github doesn't seem to offer a type for that field:

Screenshot 2022-12-09 at 10 33 41

I can define our own.

src/types.ts
marcos-arranz: I am not able to find that type. Can you point me to the definition?

src/types.ts
marcos-arranz: It is not exactly the same, `CheckResolution` can be `ERROR` or `FAILURE`, the former means that there was an error due to the application, and the later that the input given by the user was wrong. This result in one saying something like "There was an unhandled error" and other saying something like "There is an error in your yaml, please fix it". On both of them `conclusion` is `failure`

I can try to use just conclusion and report it as `action_required` for failure. WDYT?

test/app.test.ts
marcos-arranz: No, it does not explode without this.

marcos-arranz:
@dpotyralski @andersfischernielsen I have included the changes from the other PR and addressed your comments. Let me know if you have any other concern, otherwise I can merge this.

src/core-validation.ts
marcos-arranz: Addressed

src/core-validation.ts
```diff
@@ -0,0 +1,147 @@
...
+
+      const onlyChangesConfiguration = filesChanged.length === 1 && filesChanged[0] === configFileName
+
+      if (errors.length > 0) {
+        const changeRequestId = await requestPullRequestChanges(repository, prNumber, checkId)
+        log.debug(`Requested changes for PR #%d in %s.`, prNumber, changeRequestId)
+      } else if (onlyChangesConfiguration) {
+        const approvedReviewId = await approvePullRequestChanges(repository, prNumber)
+        log.debug(`Approved PR #%d in %s.`, prNumber, approvedReviewId)
+      }
```
andersfischernielsen, line 118: This was patched out on `main` 😬

andersfischernielsen: Specifically #85 and #87

PR #81, opened 2022-12-06
Prefer Either Monad to exception handling

andersfischernielsen:
This will avoid exception throwing and handling for passing information relating to templates across the application logic, since exceptions shouldn't be used for data passing and application logic.

This is preferable since we avoid any unhandled exceptions accidentally slipping through to the top-level application logic. 
This also makes it possible to reason about the types and data flow of the application and makes it clear where information is passed to/from in the validation logic.

This is accomplished using the `fp-ts` library and its `Either` Monad.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -6,7 +6,10 @@ import { git } from './git'
 import { schemaValidator } from './schema-validator'
 import { checks } from './checks'
 import 'dotenv/config'
-import { OctokitInstance, TemplateConfig, ValidationError, VersionNotFoundError } from './types'
+import { OctokitInstance, TemplateConfig, ValidationError } from './types'
+import { pipe } from 'fp-ts/function'
+import { Either, left } from 'fp-ts/lib/Either'
+import { map, separate } from 'fp-ts/lib/Array'
 
 const configFileName = process.env['TEMPLATE_FILE_PATH'] ? process.env['TEMPLATE_FILE_PATH'] : '.github/templates.yaml'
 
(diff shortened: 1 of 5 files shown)


andersfischernielsen:
Tests currently fail since I haven't updated the destructuring happening there, leading to properties being `undefined`. Just want to get initial structure feedback before I change them.

src/app.ts
```diff
@@ -125,18 +126,8 @@ const processPullRequest = async (payload: PullRequestEvent, context: Context<'p
...
-
-  const onlyChangesConfiguration = filesChanged.length === 1 && filesChanged[0] === configFileName
-
-  if (errors.length > 0) {
-    const changeRequestId = await requestPullRequestChanges(repository, prNumber, checkId)
-    log.debug(`Requested changes for PR #%d in %s.`, prNumber, changeRequestId)
-  } else if (onlyChangesConfiguration) {
-    const approvedReviewId = await approvePullRequestChanges(repository, prNumber)
-    log.debug(`Approved PR #%d in %s.`, prNumber, approvedReviewId)
-  }
```
andersfischernielsen, line 138: Note that the change request logic has been ripped out here

andersfischernielsen:
The purity of the data flow can be improved using the `TaskEither` Monad for handling async calls. That leads to everything in the main application logic being one big 

``` typescript
return pipe(
  firstOperation, 
  secondOperation,
  ...
  NthOperation
) //The return type here being Either<ValidationErrors[], ...>

andersfischernielsen:
Hmmmm, thinking a bit more about this I feel this is overkill. Simplifying the types to simple objects and having a simple combinator helper would do the trick too. 

Will refactor

andersfischernielsen:
Closing and reworking

PR #82, opened 2022-12-06
Simplify/clean up 'schema-validator' implementation

andersfischernielsen:
This simplifies and refactors the `schema-validator` implementation. 

For loops have been reduced, comments have been formatted, and type cast bugs have been fixed. 
Some minor performance improvements were gained, too 🎉


jsfr (approved):

src/schema-validator.ts
```diff
@@ -1,118 +1,99 @@
...
-  for (let index = 1; index < lines.length; index++) {
-    const newLineOffset = lines[index]
-    if (newLineOffset > offset) {
-      return index
-    }
-  }
-  return lines.length
+interface TemplateConfig {
+  repositoryConfiguration: RepositoryConfiguration
+  cstYamlRepresentation: ConcreteSyntaxTree
```
marcos-arranz, line 20: Why not keeping types in `types.ts`? If we move it here it is needed to update the references to this type in other files and export it.

src/schema-validator.ts
```diff
@@ -1,118 +1,99 @@
...
-      if (isNaN(num)) {
-        if (!CST.isScalar(item.value)) return
-
-        const key = item.key as SourceToken
-
-        if (!key) return
-
-        // If note key is not the current path value skip this node and its childs and go to next sibling
-        if (key.source !== currentPathValue) {
+      const pathValueNumber = parseInt(currentPathValue)
```
marcos-arranz, line 40: I think it is more secure to use `Number` than `parseInt` according to https://stackoverflow.com/questions/4090518/what-is-the-difference-between-parseint-and-number

PR #84, opened 2022-12-08
Broken yaml bug

marcos-arranz:
Split from https://github.com/pleo-oss/file-distributor/pull/80 

This just deals with the YAML error


src/app.ts
```diff
@@ -170,9 +191,10 @@ const processPushEvent = async (payload: PushEvent, context: Context<'push'>) =>
   const filesChanged = await getCommitFiles(repository, payload.after)
   if (!filesChanged.includes(configFileName)) return
 
-  const parsed = await determineConfigurationChanges(configFileName, repository, payload.after)
+  const either = await determineConfigurationChanges(configFileName, repository, payload.after)
 
-  if (!parsed.repositoryConfiguration) return
+  if (either.isLeft()) return
```
dpotyralski, line 196: Should we consider some debug log message here?

marcos-arranz: Makes sense. Added

.gitignore
```diff
@@ -7,6 +7,7 @@ npm-debug.log
 .env
 coverage
 lib
+.yarn
```
andersfischernielsen, line 10: What's this for? 🤔

package.json
```diff
@@ -55,5 +56,9 @@
     "smee-client": "1.2.3",
     "ts-jest": "29.0.3",
     "typescript": "4.8.4"
+  },
+  "volta": {
+    "node": "18.12.1",
+    "yarn": "1.22.19"
   }
```
andersfischernielsen, line 63: Did we ever come to a decision on whether to include/support Volta or not? 🤔

src/app.ts
```diff
@@ -53,45 +56,63 @@ const extractPullRequestInformation = (payload: PullRequestEvent) => {
...
+      const validatedTemplates = validateTemplateConfiguration(
+        {
+          repositoryConfiguration: combined,
+          cstYamlRepresentation: configurationChanges.cstYamlRepresentation,
+        },
+        mergeSchemaToDefault(defaultValueSchema),
+      )
+      const validatedFiles = validateFiles(combined, files)
+
+      return validatedTemplates.errors.concat(validatedFiles.errors)
```
andersfischernielsen, line 86: Shouldn't this be 

```suggestion
      return E.right(validatedTemplates.errors.concat(validatedFiles.errors))
```

in order to match the return type? 🤔  Or am I completely off? 😄

src/app.ts
```diff
@@ -53,45 +56,63 @@ const extractPullRequestInformation = (payload: PullRequestEvent) => {
...
+    } catch (error) {
+      if (error instanceof VersionNotFoundError) {
+        log.debug('Version for %s/%s:%s has not been found', error.owner, error.name, error.version)
+        const validatedTemplate = validateTemplateConfiguration(configurationChanges, getDefaultSchema())
+        return validatedTemplate.errors.concat({
+          message: `Version could not be found at ${error.owner}/${error.repo}:${error.version}`,
+          line: undefined,
+        })
```
andersfischernielsen, line 94: The same question here: Does this match the return type? 🤔

src/app.ts
```diff
@@ -53,45 +56,63 @@ const extractPullRequestInformation = (payload: PullRequestEvent) => {
...
+    E.match(
+      failure =>
+        Promise.resolve([
+          {
+            message: failure.message,
+            line: failure.linePos?.[0].line,
+          } as ValidationError,
+        ]),
+      success => {
+        return validateCorrectYamlChanges(success)
```
andersfischernielsen, line 112: Ah, the wrapping into `Either` happens here, right?

src/app.ts
```diff
@@ -170,9 +191,13 @@ const processPushEvent = async (payload: PushEvent, context: Context<'push'>) =>
   const filesChanged = await getCommitFiles(repository, payload.after)
   if (!filesChanged.includes(configFileName)) return
 
-  const parsed = await determineConfigurationChanges(configFileName, repository, payload.after)
+  const either = await determineConfigurationChanges(configFileName, repository, payload.after)
```
andersfischernielsen, line 194:
```suggestion
  const parsed = await determineConfigurationChanges(configFileName, repository, payload.after)
```

src/app.ts
```diff
@@ -170,9 +191,13 @@ const processPushEvent = async (payload: PushEvent, context: Context<'push'>) =>
...
 
-  if (!parsed.repositoryConfiguration) return
+  if (E.isLeft(either)) {
+    log.error('There has been an error while processing event %o', either.left)
+    return
+  }
+  const parsed = either.right
```
andersfischernielsen, line 200:
```suggestion
  if (E.isLeft(parsed)) {
    return parsed
  }
```

Coukdn't we simplify like this? 😄  Is there a need to error log, doesn't that already happen? Do we need to error log for validation errors?

src/configuration.ts
```diff
@@ -33,7 +34,11 @@ export const ensurePathConfiguration = (files?: (PathConfiguration | string)[])
 }
 
 export const configuration = (log: Logger, octokit: Pick<OctokitInstance, 'repos'>) => {
-  const determineConfigurationChanges = async (fileName: string, repository: RepositoryDetails, sha: string) => {
+  const determineConfigurationChanges = async (
+    fileName: string,
+    repository: RepositoryDetails,
+    sha: string,
+  ): Promise<E.Either<YAMLParseError, TemplateConfig>> => {
```
andersfischernielsen, line 41: This is technically a `TaskEither`, but I'm easy 🤷🏼

andersfischernielsen (approved):
Some questions and suggestions, but overall LGTM

dpotyralski (approved):
LGTM

.gitignore
marcos-arranz: Yarn 2 cache

package.json
marcos-arranz: The thing is that it is the way recommended way to setup `grim-reaper` as per https://github.com/pleo-io/grim-reaper/#setup, so if Volta is installed and then this project is used it will use Yarn 3 which means several changes that add files into the project. I haven't found any other way to prevent this than adding it to `package.json`

What is your concern adding it to `package.json`?

src/app.ts
marcos-arranz: Not exactly, `validateChanges` function returns an array of validation errors ready to be used in the check. In order to generate those errors it needs to receive a `configurationChangesOrError` which is the YAML configuration or if it is a non-valid YAML error. This is represented as `Either` and it is generated on the function `determineConfigurationChanges`

src/app.ts
marcos-arranz: This returns an array of validation errors. We can evaluate to turn everything to `Either` but I feel it falls out of the scope of this PR. WDYT?

src/app.ts
marcos-arranz: `parsed` cannot be used, as it is used on line 200. I have changed to `parsedOrError` WDYT?

src/app.ts
marcos-arranz: In this case I think so, because this happens on `processPushEvent`. It would be good to have some logging to know why `processPushEvent` exit, specially since this should never happen.

src/configuration.ts
marcos-arranz: yeah, that's nicer, but will need a major refactoring as it cannot be used `async` and `await` anymore, so it means returning `TaskEither` also from `octokit` call which in turn means also modifying the throwing part. I'd postpone to a different PR. WDYT?

PR #88, opened 2022-12-23
Reorder/simplify main application module

andersfischernielsen:
This is the first step in refactoring and simplifying the structure of the application. This PR focuses on moving logic out of the main `app.ts` module and decoupling the logic as much as currently possible by splitting it into separate modules with limited scope.

These changes will be used for the changes in the coming PRs for restructuring and simplifying the implementation of the other modules.

---

This will: 
- Rename and simplify validation functions and ensure the module follows the design of the other modules
- Remove an invalid types in the check module
- Move most logic from the app module to individual modules with limited scopes and types
- Rename `[T|t]emplate(s?)` to `[F|f]ile(s?)` globally


src/template-creation.ts
```diff
@@ -0,0 +1,39 @@
+import * as E from 'fp-ts/Either'
+import { Logger } from 'probot'
+import { configuration } from './configuration'
+import { git } from './git'
+import { templates } from './templates'
+import { OctokitInstance, RepositoryDetails } from './types'
+
+export const templateCreation = (log: Logger, octokit: Pick<OctokitInstance, 'pulls' | 'repos' | 'git' | 'issues'>) => {
```
andersfischernielsen, line 8: I lack a better name for this. The function is in charge of rendering and pushing rendered files to a repository based on the new repository configuration found on the default branch. Suggestions are very welcome 😄

andersfischernielsen: `createRepositoryFiles`? `submitRepositoryFiles`?

dpotyralski: Maybe instead of `repositoryFiles` we could go with `workflows` and have something like this: 
* `renderAndSubmitWorkflows` 
* `submitWorkflows`

or from templates POV: 
* `processTemplates`

dpotyralski (approved):

marcos-arranz (approved):

src/template-creation.ts
jsfr: The bot doesn't only process workflows though, so files/templates is more correct here than workflow.

jsfr (approved):

PR #91, opened 2022-12-27
Prefer passing data via Either instead of throwing custom exceptions

andersfischernielsen:
- Rename, simplify validation functions, module
- Remove invalid types in check module
- Move functions from app module to individual modules
- Rename coreValidation to validation
- Ensure environment importing
- Add missing linebreak
- Simplify module, narrow scopes
- Use simplified types for testing
- Rename functions, types
- Move GitHub calls to git module
- Prefer Either to throwing custom errors for passing data
- Use modified return types in tests


andersfischernielsen:
Closing in favor of a new branch.

PR #92, opened 2022-12-27
Prefer passing data via Either instead of throwing custom exceptions

andersfischernielsen:
- Rename, simplify validation functions, module
- Remove invalid types in check module
- Move functions from app module to individual modules
- Rename coreValidation to validation
- Ensure environment importing
- Add missing linebreak
- Simplify module, narrow scopes
- Use simplified types for testing
- Rename functions, types
- Move GitHub calls to git module
- Prefer Either to throwing custom errors for passing data
- Use modified return types in tests


src/templates.ts
```diff
@@ -87,20 +92,8 @@ export const templates = (log: Logger, octokit: Pick<OctokitInstance, 'repos'>)
...
-        throw new VersionNotFoundError(
-          error.message,
-          templateRepository.owner,
-          templateRepository.repo,
-          templateVersion,
-        )
-      }
-      throw error
-    }
+    const release = await getReleaseFromTag(templateVersion, templateRepository.owner, templateRepository.repo)
```
marcos-arranz, line 95: We should mantein the check 
```
if (error instanceof RequestError && error.status === 404)
```
on call `getReleaseFromTag` otherwise, any error like a 500, or a network error will be reported as a Version error

src/validation.ts
```diff
@@ -124,18 +93,6 @@ export const validation = (
...
-        {
-          ...checkInput,
-          conclusion: 'failure',
-          checkRunId: previousCheckId,
-          errors: [],
-        },
-        configFileName,
-      )
-      await updateCheck(resolved)
-      throw error
```
marcos-arranz, line 138: We still need to capture any unhandled error and update the check accordingly. Otherwise, the check would be left in a `pending` state

jsfr (approved):
Good with me (apart from the comments from Marcos, which need to be addressed).

src/configuration.ts
```diff
@@ -50,46 +56,41 @@ export const configuration = (log: Logger, octokit: Pick<OctokitInstance, 'repos
...
+        defaultBranch: repository.defaultBranch,
+      },
+    }
 
-      const lineCounter = new LineCounter()
+    log.debug('Saw combined configuration contents %o', combinedConfiguration)
 
-      const cst = new Parser(lineCounter.addNewLine).parse(decodedContent)
-      const tokens = Array.from(cst)
+    return E.right(combinedConfiguration)
```
dpotyralski, line 77: Nice :)

src/configuration.ts
```diff
@@ -50,46 +56,41 @@ export const configuration = (log: Logger, octokit: Pick<OctokitInstance, 'repos
     const decodedContent = Buffer.from(content, 'base64').toString()
     log.debug('%s contains: %s', fileName, decodedContent)
 
-    try {
-      const parsed: RepositoryConfiguration = parse(decodedContent)
-      log.debug('Saw configuration file contents %o', parsed)
+    const contents = parseDocument(decodedContent)
+    if (contents.errors.length > 0) return E.left(contents.errors)
```
dpotyralski, line 60: 🎉

src/templates.ts
andersfischernielsen: Whether it's a 404 or anything else, we still can't validate the version. Currently the error will report that we couldn't fetch the release, which indicates that either permissions are set up incorrectly for the application or that the version isn't available 😄

dpotyralski (approved):
LGTM!

src/templates.ts
marcos-arranz: From my pov, it is valuable to the user to specify if the error comes from a missing version or is a transient/unknown error.

PR #94, opened 2023-01-02
Update Overview.md

r0binary:
I fixed a typo the I spotted when reading through this. Thanks for the nice documentation.

### DIFF
--- a/docs/Overview.md
+++ b/docs/Overview.md
@@ -3,7 +3,7 @@
 ## GitHub Webhooks
 GitHub Apps can listen to webhooks being emitted based on what's happening in the organisation they're installed into. These webhook events are configured in the organisation page in GitHub. 
 
-This app depends on one event type: `push`. The app subscribed to all `push` events happening on any repository within the organisation and will determine what to do based on the even `payload`. 
+This app depends on one event type: `push`. The app subscribed to all `push` events happening on any repository within the organisation and will determine what to do based on the event `payload`. 
 
 ## Flow
 The app will continue execution if the `push` event happens on any of the configured branches to listen for changes on. If the `push` event happened on any of these branches, the file changes in the given commit leading to the `push` event being emitted are examined. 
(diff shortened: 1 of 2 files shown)


docs/Overview.md
```diff
@@ -3,7 +3,7 @@
 ## GitHub Webhooks
 GitHub Apps can listen to webhooks being emitted based on what's happening in the organisation they're installed into. These webhook events are configured in the organisation page in GitHub. 
 
-This app depends on one event type: `push`. The app subscribed to all `push` events happening on any repository within the organisation and will determine what to do based on the even `payload`. 
+This app depends on one event type: `push`. The app subscribed to all `push` events happening on any repository within the organisation and will determine what to do based on the event `payload`.
```
andersfischernielsen, line 6:
```suggestion
This app depends on one event type: `push`. The app subscribes to all `push` events happening on any repository within the organisation and will determine what to do based on the event `payload`. 
```

andersfischernielsen (approved):
Thanks! 😄

jsfr (approved):

andersfischernielsen:
@r0binary I think you need to create this branch on the repository itself in order to allow CI to run as expected 🤔

r0binary:
@andersfischernielsen Do you think that makes a difference? \i was wondering if the kinda weird branch name could have broken it. WDYT?

r0binary:
@andersfischernielsen I created a clone of my branch with a different name. This seems to make no difference. Also I tried to create a branch in this very repo which I seem to not have the permissions for. What Github did when I created the fix was to fork the repo, create a branch in the fork and create a PR from the branch in the fork to pleo-oss:main.
I think this is the common way to do this in OSS projects. So we should see how the CI tooling can be fixed to handle this properly.

dpotyralski (approved):

r0binary:
I added some more triggers to get test workflow, because it's results are expected to be present at some time, but due to the fact, that there never was a feature branch in the repo the on_push event did not fire and the workflow was waiting for the results forever. So i added the same on_pullrequest triggers that verify the release.

PR #97, opened 2023-01-09
Use check outcome for logging

andersfischernielsen:
This fixes an `[Object object]` being logged after check creation/update in the logs and will instead use the `conclusion` property for useful logging.


dpotyralski (approved):
LGTM!

jsfr (approved):

PR #102, opened 2023-01-11
Log application version on startup

andersfischernielsen:
This will log the application version on startup to validate that the expected version is running when hosting the application container image.


src/start.ts
```diff
@@ -80,6 +81,8 @@ if (!appId || !privateKey) {
   }
 }
 
+log.info(`Running Application v${version}`)
```
jsfr, line 84:
```suggestion
log.info(`Running file-distributor v${version}`)
```

jsfr (approved):

dpotyralski (approved):

PR #105, opened 2023-01-11
Ensure latest changes to the default branch are being included in released container images

andersfischernielsen:
When the application is released, a `CHANGELOG` update and version bump to `package.json` are pushed to the default branch by our release tooling. These changes are currently not included in the released container image, leading to the "internal" version in the container image being incorrect. 

This fixes that by using the current state of the default branch for release. With the current concurrency policy of the release workflow we avoid including changes from other PRs if these happen on the default branch at the same time.


dpotyralski:
> This fixes that by using the current state of the default branch for release. With the current concurrency policy of the release workflow we avoid including changes from other PRs if these happen on the default branch at the same time.

So without `ref: ${{ github.event.repository.default_branch }}` it wasn't taking the current default branch latest commit?

andersfischernielsen:
> > This fixes that by using the current state of the default branch for release. With the current concurrency policy of the release workflow we avoid including changes from other PRs if these happen on the default branch at the same time.
> 
> So without `ref: ${{ github.event.repository.default_branch }}` it wasn't taking the current default branch latest commit?

@dpotyralski Nope, it was using the `ref` that the `push` event happened on - in this case the state of the `main` branch before we push the version update file changes 😄

dpotyralski (approved):

andersfischernielsen:
> Does grim-reaper suffer from this potential issue as well?

@jsfr [Nope](https://github.com/pleo-io/grim-reaper/blob/a059d3fe8804b4a7533c318a6c7c8a9257eb16da/.github/workflows/release.yaml#L70) 👍🏼

PR #117, opened 2023-01-18
Fix wrong line report and simplify implementation

marcos-arranz:
This fixes the wrong line reporting bug.

The error was due to parsing and then creating the CST from the stringified parsed version which result in losing information about lines and comments. In order to fix it the parsing is moved to the validation, which parses retaining the source tokens with the option `{ keepSourceTokens: true }`. The custom tree search is not needed as YAML library already provides the method `getIn` on the `Document.Parsed<ParsedNode>` which returns the node given a path.


src/schema-validator.ts
```diff
@@ -123,24 +75,24 @@ export const schemaValidator = (log: Logger) => {
...
 
   const validateConfiguration = (
-    configuration: RepositoryConfiguration,
+    parsedDocument: Document.Parsed<ParsedNode>,
     schema: Schema,
-    cstRepresentation: CSTRepresentation,
+    lineCounter: LineCounter,
   ): ConfigurationValidation => {
-    const validateConfiguration = ajv.compile<RepositoryConfiguration>(schema)
+    const validateConfigurationFn = ajv.compile<RepositoryConfiguration>(schema)
```
andersfischernielsen, line 82: I'd keep the original naming for this functions, since it's obvious it's a function based on the type of the const - WDYT?

src/types.ts
```diff
@@ -8,11 +7,6 @@ export interface PathConfiguration {
 
 export type ConfigurationValues = { [key: string]: string | undefined }
 
-export interface CSTRepresentation {
-  tokens: Token[]
-  lines: number[]
-}
```
andersfischernielsen, line 14: I removed this, since it's never used anywhere now

src/schema-validator.ts
```diff
@@ -1,82 +1,34 @@
...
+
+  const optionalNode = O.fromNullable(parsedDocument.getIn(paths, true) as Scalar)
+
+  return pipe(
+    optionalNode,
+    O.map(node => O.fromNullable(node.srcToken?.offset)),
+    O.flatten,
+    O.map(n => lineCounter.linePos(n).line),
+    O.toUndefined,
+  )
```
andersfischernielsen, line 31: This can be simplified to: 

```suggestion
  const optionalNode = parsedDocument.getIn(paths, true) as Scalar
  const offset = optionalNode?.srcToken?.offset ?? -1
  if (offset !== -1) lineCounter.linePos(offset).line
  return undefined
```

WDYT? Then we can also remove the use of `fp-ts`'s `Optional` here, since optionals are built into the language 😄

src/schema-validator.ts
andersfischernielsen: If it has to be really obvious, the type can also be explicitly added like so: 

```suggestion
    const validateConfiguration: ValidateFunction<RepositoryConfiguration> = ajv.compile<RepositoryConfiguration>(schema)
```

😄

src/types.ts
marcos-arranz: Nice catch!

src/schema-validator.ts
marcos-arranz: It's fine I can revert the change back, actually it is callable but also has attributes so it's kind of not a function 🫤

src/schema-validator.ts
marcos-arranz: Since we want to go more functional I think I will keep the code, if you don't mind. Although there are nullable types on TS,  using `Optional` lets using `map` and other functional functions, so there is a chain of functions applied.

src/git.ts
```diff
@@ -344,6 +344,22 @@ export const git = (log: Logger, octokit: Pick<OctokitInstance, 'pulls' | 'repos
...
+    const { data: fileContents } = await octokit.repos.getContent({
+      ...repository,
+      path: fileName,
+      ref: sha,
+      request: {
+        retries: 3,
+        retryAfter: 1,
+      },
+    })
+    const { content } = fileContents as { content: string }
```
jsfr, line 357: Just a heads-up, but this is fallible. `content` is only present if the file is below a certain size. This is why grim-reaper use a check for `if (!('content' in data)) { return/throw }` instead, which removes the need for casting and handles the situation gracefully as well.

jsfr (approved):

src/git.ts
marcos-arranz: Makes sense, I think we can have a different PR to deal with errors with `TaskEither` and `Either`

src/schema-validator.ts
andersfischernielsen: Nice 😄 

Yeah, true that's a bit strange 🤔 They do call it a function [in the docs](https://ajv.js.org/api.html#ajv-compile-schema-object-data-any-boolean-promise-any), though

src/schema-validator.ts
andersfischernielsen: My point was mainly that: 
- we use the language builtins
- the code is shorter/simpler (less intermediary steps) and therefore easier to reason about for the reader

But I'm fine with keeping it in 😄

andersfischernielsen (commented):
What happened to the `yarn.lock`? 🤔

andersfischernielsen (approved):

PR #121, opened 2023-01-25
Present line numbers for invalid file specifications in configurations

andersfischernielsen:
This uses the present line numbers in the file validation if present. Some minor cleanup on the validation was also done as part of this. 

This was done as part of investigating whether the line detection in configuration errors works as expected when default values are injected. I haven't been able to find any issues when additional default values are injected into the configuration. 

---

Tested extensively in https://github.com/andersfischernielsen/test-template-configuration/pull/20


src/checks.ts
```diff
@@ -27,9 +27,11 @@ const withErrors = (
   errors: ValidationError[],
   configFilePath: string | undefined,
 ): RestEndpointMethodTypes['checks']['update']['parameters'] => {
-  const withoutErrorLine = errors.filter(e => !e.line)
-  const withErrorLine = errors.filter(e => e.line)
-  const text = withoutErrorLine.length > 0 ? withoutErrorLine.map(e => `- \`${e.message}\``).join('\n') : undefined
+  const [withLines, withoutLines] = errors.reduce<[ValidationError[], ValidationError[]]>(
+    ([p, m], e) => (e.line !== undefined ? [[...p, e], m] : [p, [...m, e]]),
```
andersfischernielsen, line 31: `e.line !== undefined` fixes a logic bug when `line === 0`. This would lead to errors on `line: 0` not being reported on line 0.

andersfischernielsen: This also does a single pass through the list to not have to do two iterations in the previous double `filter`.

andersfischernielsen:
I set up auto-formatting in CI which follows the ESLint configuration. I hope this doesn't make the small number of changes too hard to review 😄

package.json
```diff
@@ -16,8 +16,8 @@
     "build:run": "yarn build && yarn start",
     "start": "probot run ./lib/app.js",
     "test": "jest",
-    "lint": "prettier --check src/**/*.ts test/**/*.ts && eslint --ext .ts src && eslint --ext .ts test",
```
marcos-arranz, line 19: Why to move to `eslint`?

src/checks.ts
marcos-arranz: Nice catch on the `0` error. 

Regarding the `reduce` function, from my point of view it's a smart trick, but a bit hard to read compared with having two filters, but I am easy.

src/schema-validator.ts
```diff
@@ -47,24 +47,37 @@ export const mergeSchemaToDefault = (valuesSchema: Schema) => {
...
+  const errors = paths?.reduce<Set<[string, number | undefined]>>(
     (errors, file) =>
       files.some(t => new RegExp(file.source).test(t))
         ? errors
-        : errors.add(
+        : errors.add([
             `'${file.source}' was not found in the release '${version}' - make sure the file is present in the release or delete the file from your configuration.`,
-          ),
-    new Set<string>(),
+            withLines?.get(file.destination),
```
marcos-arranz, line 71: Nice!

package.json
andersfischernielsen: We're not moving to ESLint, it's always been part of the project. This just ensures ESLint fixes code issues before Prettier formats the code according to style guides.

andersfischernielsen: If we don't do this, then you'd need to run the `format` task twice in succession, which is silly 😄

EliasJorgensen (approved):

PR #124, opened 2023-02-01
Migrate from yarn to pnpm

andersfischernielsen:
This migrates the project from `yarn` to `pnpm`. 

All instances of `yarn` tasks have been migrated to `pnpm` as well as relevant `.gitignore`s, `Dockerfile`s, etc.

Tested locally both via running the app through `pnpm` as well as through `docker`.


src/app.ts
```diff
@@ -73,4 +74,6 @@ export default async (app: Probot) => {
   app.on(['pull_request.opened', 'pull_request.synchronize'], async (context: Context<'pull_request'>) => {
     await processPullRequest(context.payload as PullRequestEvent, context)
   })
+
+  app.log.info(`Running file-distributor v${version}.`)
```
andersfischernielsen, line 78: This was lost as part of #114

src/app.ts
```diff
@@ -57,6 +58,8 @@ const processPushEvent = async (payload: PushEvent, context: Context<'push'>) =>
 }
 
 export default async (app: Probot) => {
+  app.log.info(`Running file-distributor v${version}`)
```
andersfischernielsen, line 61: This was lost as part of #114

Dockerfile
```diff
@@ -2,21 +2,24 @@ FROM node:19-alpine AS BUILD_IMAGE
 
 WORKDIR /usr/src/app
 
-COPY yarn.lock package.json tsconfig.json .
+COPY pnpm-lock.yaml package.json tsconfig.json .
 COPY src/ ./src
 COPY .swcrc ./.swcrc
 
-RUN yarn && yarn build && rm -rf node_modules && yarn --production
+RUN npm install -g pnpm
```
jsfr, line 9:
```suggestion
RUN npm install -g pnpm@7
```

Dockerfile
```diff
@@ -2,21 +2,24 @@ FROM node:19-alpine AS BUILD_IMAGE
...
+RUN pnpm install && pnpm build && rm -rf node_modules && pnpm install --prod
 
 ###################
 
 FROM node:19-alpine
 
+RUN npm install -g pnpm
```
jsfr, line 16:
```suggestion
RUN npm install -g pnpm@7
```

Dockerfile
```diff
@@ -2,21 +2,24 @@ FROM node:19-alpine AS BUILD_IMAGE
...
 WORKDIR /usr/src/app
 
-COPY yarn.lock package.json tsconfig.json .
+COPY pnpm-lock.yaml package.json tsconfig.json .
 COPY src/ ./src
 COPY .swcrc ./.swcrc
 
-RUN yarn && yarn build && rm -rf node_modules && yarn --production
+RUN npm install -g pnpm
+RUN pnpm install && pnpm build && rm -rf node_modules && pnpm install --prod
```
jsfr, line 10:
```suggestion
RUN pnpm install --frozen-lockfile \
  && pnpm build \
  && pnpm install --frozen-lockfile --prod
```

Frozen lockfile ensure nothing gets changed from what is specified. Also running `pnpm` with `--prod` will automatically remove dev modules from node_modules, so it's not a requirement to delete it first.

jsfr (approved):

Dockerfile
andersfischernielsen: Oooh, that's a nice bonus! Nice catch 🔥

Dockerfile
```diff
@@ -2,21 +2,26 @@ FROM node:19-alpine AS BUILD_IMAGE
 
 WORKDIR /usr/src/app
 
-COPY yarn.lock package.json tsconfig.json .
+COPY pnpm-lock.yaml package.json tsconfig.json .
```
dpotyralski, line 5:
```suggestion
COPY pnpm-lock.yaml package.json tsconfig.json ./
```

dpotyralski (approved):

Dockerfile
andersfischernielsen: Isn't this redundant? I don't think this will lead to a functional change :smile: Especially since it's unchanged 🤔

dpotyralski: I've taken that suggestion from my Intellj: `Destination must be a directory and end with a ''/'' ` and added it here as well. All works without it, that's why you have your PR approved anyway :)

PR #128, opened 2023-02-02
Revert release behavior to previous 'npx'-based flow

andersfischernielsen:
This will ensure the release of the application as previously seen in the repository. 

When the shared `.autorc.json` was used for releasing the repository via Auto, the `git-tag` plugin was used instead of the previously used `npm` default plugin. This lead to `package.json` not being updated as expected. 

This is fixed by not using the shared `.autorc.json` and instead relying on a near-identical configuration, which mimics the default behavior of invoking Auto via `npx` (as the repository had previously) while preserving the `plugins` found in the shared `.autorc.json`.

---

This also fixes some unrelated missing type errors in the application implementation. Technically not related to releasing, but I hope that's okay.


src/release.ts
```diff
@@ -29,7 +29,11 @@ export const release = (log: Logger, octokit: Pick<OctokitInstance, 'repos'>) =>
     if (picked) log.debug('Using %s for %s.', picked.name, source)
 
     const text = (await picked?.async('text')) ?? ''
-    return text?.replace(/#<<</gm, '<<<')
+    return text
+      ?.replace(/#<<</gm, '<<<')
+      ?.replace(/#>>>/gm, '>>>')
+      ?.replace(/\/\/<<</gm, '<<<')
+      ?.replace(/\/\/>>>/gm, '>>>')
```
marcos-arranz, line 36: Is this related with the original problem?

andersfischernielsen: It's mentioned in the description of the PR :smile:

marcos-arranz: sorry, I missed it. This isn't a type error in though, I understand this adding missing substitutions for `>>>` and so on. Without these changes how could this work?

andersfischernielsen: We already have these substitutions. This adds an additional one for `//`.

andersfischernielsen: It's a two-liner to support `//` on top of the existing substitution for `#` - the rest is auto-formatting

jsfr (approved):

dpotyralski (approved):

PR #131, opened 2023-02-17
Validate containerization on top of build steps

andersfischernielsen:
This will 
- Fix Dockerfile issues
- Verify container building in CI checks (we need to know that our container builds for releasing)
- Cache containerization since it's faster (container builds only take [a few seconds](https://github.com/pleo-io/file-distributor/actions/runs/4201443440/jobs/7288451744#step:5))
- Run validation/formatting steps in parallel since it's faster (we have no inter-job dependencies)


Dockerfile
```diff
@@ -2,9 +2,8 @@ FROM node:19-alpine AS BUILD_IMAGE
 
 WORKDIR /usr/src/app
 
-COPY pnpm-lock.yaml package.json tsconfig.json .
+COPY pnpm-lock.yaml package.json tsconfig.json .swcrc ./
```
andersfischernielsen, line 5: You were right @dpotyralski 😄 `docker` reports this as an error too, now

jsfr (approved):

dpotyralski (approved):
Sweet! 🧁

PR #132, opened 2023-03-06
Add 'environment' property to Release CI job (to start tracking "GH deployments")

VictorPascualV:
- Add 'environment' property to Release CI job (to start tracking "GH deployments")


.github/workflows/release.yaml
```diff
@@ -35,6 +35,7 @@ jobs:
     name: Push Docker image to Docker Hub
     needs: release
     runs-on: ubuntu-22.04
+    environment: release
```
LittleGreenYoda42, line 38: Is the docker image push the deployment or the NPM release above? Where should the deployment be indicated?

LittleGreenYoda42 (approved):

.github/workflows/release.yaml
andersfischernielsen: Well, it's all kinda a release. We release to GitHub above mainly to generate a CHANGELOG update and publish the Git Tag. We still deploy `file-distributor` manually by bumping the docker tag in `terraform`, so technically none of this is a deploy, just a release

andersfischernielsen: But once it's in GHCR we _can_ deploy it, so I guess this is the closest step to an actual deployment as we get in this workflow

andersfischernielsen (approved):
🍰

.github/workflows/release.yaml
VictorPascualV: @andersfischernielsen for now I'll merge as is, but please feel free to adjust "what is a deploy/release" for your team in the future.

andersfischernielsen: Sure thing! One day (TM) all of this will be triggered automatically

PR #140, opened 2023-05-25
Update opslevel.yml [skip ci]

jsfr:
Update the owner of the component

---

## Release Notes

<A description of the changes made for inclusion in the auto-generated CHANGELOG.md>


andersfischernielsen (commented):
What's with the 

```
opslevel.yml#L16

must be string
```
? 🤔 

Any way to get the current expected value? Can it be any string? Specific strings?

jsfr:
> What's with the
> 
> ```
> opslevel.yml#L16
> 
> must be string
> ```
> 
> ? 🤔
> 
> Any way to get the current expected value? Can it be any string? Specific strings?

It's because we use the JSON schema from OpsLevel and in their schema tags can _only_ be strings but currently it is set to a boolean here. I think we need to change it to a string actually 🤔

andersfischernielsen (approved):

r0binary (approved):

PR #169, opened 2023-07-23
Log timings of core functionality

andersfischernielsen:


Niceplace (approved):

src/app.ts
```diff
@@ -57,6 +57,18 @@ const processPushEvent = async (payload: PushEvent, context: Context<'push'>) =>
   await pushFiles(repository, after, configFileName)
 }
 
+const logInterval = (interval: [number, number], context: Context): void => {
+  const [seconds, nanoseconds] = interval
+  const milliseconds = nanoseconds / 1000000
```
r0binary, line 62: This does not look quite right. Did you mean `const milliseconds = nanoseconds * 1000000`?

r0binary (changes requested):
There seems to be a flaw in the code. Apart from that it looks good to me

src/app.ts
r0binary: Oh damn, sorry I seem to be still a bit sleepy. Nevermind

r0binary (approved):

PR #171, opened 2023-07-25
Support merging PRs automatically if enabled in repository configurations

andersfischernielsen:
According to [this](https://github.com/orgs/community/discussions/9829), the App should be able to merge PRs as an admin via the API. I've tested this on a test repository and it seems to be the case.

---

## Release notes

Support merging PRs created as a result of configuration changes automatically if `automerge: true` is set in the repository configuration. 

This requires making the App an admin on either the repository to merge into or on the organization the App is enabled on.


src/git.ts
```diff
@@ -286,6 +286,21 @@ export const git = (log: Logger, octokit: Pick<OctokitInstance, 'pulls' | 'repos
     return id
   }
 
+  const mergePR = async (repository: RepositoryDetails, pullRequestNumber?: number) => {
+    if (!pullRequestNumber) return false
```
LittleGreenYoda42, line 290: Do we just want to return false if there is no PR number?

I would suggest to throw an error if we don't expect that to happen or at least log a warning.

This feels like a situation where we will wonder why it suddenly stopped working because there was a change and suddenly the number is undefined.

andersfischernielsen: We definitely do expect that to happen 😄 If the changes in the configuration will lead to no files changed, there's no need to create the PR - hence the PR number will be `undefined`.

danieldisu (approved):

src/git.ts
LittleGreenYoda42: I would make the pullRequestNumber mandatory and move the check for it to be defined out to make it more obvious what is happening.

src/file-creation.ts
```diff
@@ -31,6 +31,8 @@ export const fileCreation = (
     const { version, files: processed } = await renderFiles(combined)
     const pullRequestNumber = await commitFilesToPR(repository, version, processed)
 
+    if (combined.automerge) mergePR(repository, pullRequestNumber)
```
LittleGreenYoda42, line 34: I would rather add another condition to the if for checking that the pullRequestNumber is defined. That way it is more obvious that it could be undefined and nothing would happen in that case. We could also have some helper like `needsAutoMerge`.

LittleGreenYoda42 (changes requested):

src/git.ts
andersfischernielsen: Done!

LittleGreenYoda42: Thanks :+1:

LittleGreenYoda42 (approved):

src/file-creation.ts
```diff
@@ -31,6 +31,8 @@ export const fileCreation = (
     const { version, files: processed } = await renderFiles(combined)
     const pullRequestNumber = await commitFilesToPR(repository, version, processed)
 
+    if (combined.automerge && pullRequestNumber !== undefined) mergePR(repository, pullRequestNumber)
```
LittleGreenYoda42, line 34: What do you think about moving this down to after the logging has happened. Right now it say "merged pr" and after that it committed the files. Just reading the logs this could be confusing.

LittleGreenYoda42: Sorry for not bringing this up earlier. I just had another look at the code 😬

marcos-arranz (approved):

PR #176, opened 2023-07-28
Support automerging PRs iff. all CI checks pass for a given PR

andersfischernielsen:
Support waiting for CI checks on PRs to pass for up to a configured number of retries before automatically merging them.


LittleGreenYoda42 (approved):
I get the idea behind this change. I am wondering if we could have potential problems now that the execution runs for a lot longer than before. I don't know how and where this runs and if we might have a limited number of job runners.

andersfischernielsen:
> I get the idea behind this change. I am wondering if we could have potential problems now that the execution runs for a lot longer than before. I don't know how and where this runs and if we might have a limited number of job runners.

@LittleGreenYoda42 I was wondering about that myself - after a discussion with Erwin (and the observed behavior) I'm convinced this doesn't block the Node.js event loop. 

This app is designed to run on Node.js (though other runtimes are supported), so it'd depend on the configuration values for retries etc. No additional web workers etc. are spun up. 😄

andersfischernielsen:
> Some more context on how and where the file-distributor code runs: https://getpleo.slack.com/archives/C050ZHF002V/p1690544766581979

@LittleGreenYoda42 Do note that this repository is completely open-source and that any contributors outside of `pleo-io` can't see this discussion 😄

r0binary (approved):

MiguelMRoldan (approved):

PR #179, opened 2023-09-22
[DEVX-1655] - Remove log commit

marcos-arranz:
This removes the logging of a commit


dpotyralski (commented):
Shouldn't we change the [log level ](https://github.com/pleo-io/terraform/blob/911e3447d92aabf2f0a7a2bce6eea5a2e1e2566f/components/file-distributor/main.tf#L63)for file-distributor in general?

marcos-arranz:
@dpotyralski probably, but we shouldn't log the whole commit in any case

dpotyralski (approved):

just-jim (approved):
Let's make sure that there is no other `log.debug` in there! 😅  (or in other pleo repos)

PR #198, opened 2023-10-27
Use auto approver bot and add it as codeowner

marcos-arranz:
Use Pleo Bot auto approver and add it as codeowner to bypass requirements


VictorPascualV (approved):
A bit off-topic but, I was slightly surprised that his repo is open-sourced, and, therefore, public to people outside Pleo.

Let's make sure that the changes that we introduce make sense in the open-source context. Otherwise, we'll probably need to switch the repo back to private.

marcos-arranz:
> A bit off-topic but, I was slightly surprised that his repo is open-sourced, and, therefore, public to people outside Pleo.
> 
> Let's make sure that the changes that we introduce make sense in the open-source context. Otherwise, we'll probably need to switch the repo back to private.

Yeah, we may reevaluate if we want it to be public

mjio (approved):

PR #231, opened 2023-11-08
feat: use pleo wide renovate config

r0binary:
* to get the latest renovate config defaults it is recommended to use the pleo wide base config

---

## Release Notes

Internal config changes


jsfr (approved):

.github/renovate.json
```diff
@@ -1,28 +1,7 @@
...
-    "automerge": true,
-    "addLabels": [
-      "autoapprove"
-    ]
-  },
-  "platformAutomerge": true,
-  "schedule": "after 8am and before 5pm every weekday"
-}
\ No newline at end of file
+  "platformAutomerge": true
```
VictorPascualV, line 6: Why have we decided to keep this? is it different from the default config? if yes, why?

VictorPascualV (approved):

.github/renovate.json
r0binary: Yes, that's not in the default config. I assume we can get rid of it, but would like to verify with Marcos first

VictorPascualV: FYI @marcos-arranz 👆

Niceplace (changes requested):
Change requested

- Apply overrides to local config to enable [lockfileMaintenance](https://docs.renovatebot.com/configuration-options/#lockfilemaintenance), example: https://github.com/pleo-io/metrics-collector/blob/main/.github/renovate.json#L8-L13. This will ensure that the lockfile is updated when dependencies are updated 
- Validate that we are in the clear to pull renovate config (private repository) from this public repository

Niceplace (approved):

PR #283, opened 2023-11-29
Update dependencies

jsfr:
Updates all dependencies to their latest version.


MiguelMRoldan (approved):
Looks good to me

.tool-versions
```diff
@@ -0,0 +1,2 @@
+node 21.2.0
```
r0binary, line 1: Some quick googling revealed this file is used by asdf. Is that the case here too?I feel at this point asking might be faster than continued search

r0binary (approved):
Thanks @jsfr. But shouldn't this be covered by renovate?

jsfr:
> Thanks @jsfr. But shouldn't this be covered by renovate?

Not e.g. the upgrade to Node 21 in our workflow files, I believe. We were in a weird state here because we use Node 21 in the Dockerfile but not anywhere else, meaning build, local and test ran with Node 20, but the production Dockerfile ran with Node 21.

Also Renovate is slow and I need a new release with everything updated for the task this PR is related to :)

.tool-versions
jsfr: `.tool-versions` can be used by many different things, but e.g. asdf yes. Committing this file means it's easier to ensure everyone uses the same version locally, it's not a hard requirement though.

PR #300, opened 2023-12-08
Feature: improve waiting logic for PR checks to succeed before merging the PR

LittleGreenYoda42:
<Implementation details>

If the ENSURE_PASSING_PR_CHECKS env variable is enabled for the file-distributor, it will wait for the PR checks and only automatically merge the PR if they succeed. With the MAX_PR_CHECK_ATTEMPTS env variable one can set how often the file distributor will check for the checks status which is done every 10 seconds. The default number of attempts is 300 which means that the file-distributor will wait for 50 minutes for the checks (e.g. test executions) to succeed. If the checks are marked as failed, the status check will be cancelled and the PR not merged automatically. 

---

## Release Notes

Improving the logic for the file-distributor to wait for checks to succeed before automatically merging the PR to update the files. If the checks are marked as failed, the file-distributor will stop checking the status and not merge the PR. If the number of attempts to get the checks status is not defined with the `MAX_PR_CHECK_ATTEMPTS` env variable a default of 300 is used which means checking for 50 minutes.

<A description of the changes made for inclusion in the auto-generated CHANGELOG.md>


src/git.ts
```diff
@@ -298,40 +300,51 @@ export const git = (log: Logger, octokit: Pick<OctokitInstance, 'pulls' | 'repos
...
-          resolve(result)
-        }
-
-        if (count > attempts) {
-          log.debug(`Waited for %d times.`, count)
-          clearInterval(interval)
-          resolve(result)
+        switch (true) {
+          case result === CANCEL_TOKEN:
+            clearInterval(interval)
```
Niceplace, line 312: We are clearing the interval in all cases so should it really be inside of the switch statement ?

src/git.ts
```diff
@@ -298,40 +300,51 @@ export const git = (log: Logger, octokit: Pick<OctokitInstance, 'pulls' | 'repos
...
-        if (result) {
-          clearInterval(interval)
-          resolve(result)
-        }
-
-        if (count > attempts) {
-          log.debug(`Waited for %d times.`, count)
-          clearInterval(interval)
-          resolve(result)
+        switch (true) {
```
Niceplace, line 310: Would three separate `if` statements be more readable in this case ? We could simplify this logic

```
clearInterval(interval)
if (!!result) {
  resolve(result)
} else {
   resolve()
}
```

Could we ignore `case count >= attempts:` since it only logs a debug line ?

LittleGreenYoda42: I had it as three `if else` statements before and was not a huge fan of it. With that said I also don't have strong feelings either. 

Unfortunately we can't get rid of the `count >= attempts` case because it is our exit criterion for stopping the polling for checks status to not run this forever.

src/git.ts
```diff
@@ -298,40 +300,51 @@ export const git = (log: Logger, octokit: Pick<OctokitInstance, 'pulls' | 'repos
...
+          case result === CANCEL_TOKEN:
```
dpotyralski, line 311: What is "CANCEL_TOKEN" and in which cases it may happen?

src/git.ts
```diff
@@ -298,40 +300,51 @@ export const git = (log: Logger, octokit: Pick<OctokitInstance, 'pulls' | 'repos
...
     attempts: number,
     delay: number = 10000,
-  ): Promise<T> => {
+  ): Promise<T | void> => {
     let count = 0
-    return new Promise<T>(resolve => {
+    return new Promise<T | void>(resolve => {
       const interval = setInterval(async () => {
         const result = await condition()
+        count++
```
dpotyralski, line 308: {minor} Could we maybe name it more explicitly? `currentAttempt`? The same applies with `attempts` maybe `maxAttempts`?

src/git.ts
LittleGreenYoda42: The cancel token is returned from the `passesCIChecks` function in case the checks were marked as `failure` and it does not make sense to wait any longer.

See here:
https://github.com/pleo-io/file-distributor/blob/3fd81b9e94031589bbba197faf4cdcbb474961e9/src/git.ts#L338

src/git.ts
LittleGreenYoda42: Sure

src/git.ts
dpotyralski: I see, thanks for clarifications 👍

src/git.ts
```diff
@@ -349,12 +362,20 @@ export const git = (log: Logger, octokit: Pick<OctokitInstance, 'pulls' | 'repos
         ...repository,
         pull_number: pullRequestNumber,
       })
-      const times = Number(process.env['MAX_PR_CHECK_ATTEMPTS'])
+
+      let prCheckAttempts = Number(process.env['MAX_PR_CHECK_ATTEMPTS'])
+
+      if (Number.isNaN(prCheckAttempts) || prCheckAttempts <= 0) {
+        prCheckAttempts = 300
```
dpotyralski, line 369: Could we use a constant to handle this value? We have something similar here for config file:
```
const configFileName = process.env['TEMPLATE_FILE_PATH'] ? process.env['TEMPLATE_FILE_PATH'] : '.github/templates.yaml'
```

LittleGreenYoda42: What we can do is 

```
const prCheckAttempts = Number(process.env['MAX_PR_CHECK_ATTEMPTS']) || 300
```

This will try to convert the string env value to a number and if this is NaN or 0 it will set the default of 300. This does not cover the case where the value could be smaller than 0 but I could live with that.

dpotyralski: Works for me 👍

LittleGreenYoda42: Done

dpotyralski (approved):
LGTM

